PR #13, opened 2025-05-28
Use refreshable credentials for OpenSearch client

rithin-pullela-aws:
### Description
Using refreshable credentials to create the client. This ensures that temporary credentials (from EC2 instance profiles or other IAM roles) automatically refresh without requiring the application to reinitialize the client.

### Issues Resolved
https://github.com/opensearch-project/opensearch-mcp-server-py/issues/3

### DIFF
--- a/src/opensearch/client.py
+++ b/src/opensearch/client.py
@@ -58,14 +58,14 @@ def initialize_client() -> OpenSearch:
 
     # 2. Try to get credentials (boto3 session)
     try:
-        credentials = boto3.Session().get_credentials()
+        credentials=boto3.Session().get_credentials()
         if credentials:
-            aws_auth = AWS4Auth(
-                refreshable_credentials = credentials,
-                service = OPENSEARCH_SERVICE,
-                region = aws_region,
+            aws_auth=AWS4Auth(
+                refreshable_credentials=credentials,
+                service=OPENSEARCH_SERVICE,
+                region=aws_region,
             )
-            client_kwargs['http_auth'] = aws_auth
+            client_kwargs['http_auth']=aws_auth
             return OpenSearch(**client_kwargs)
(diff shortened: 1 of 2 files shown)


src/opensearch/client.py
```diff
@@ -61,11 +61,9 @@ def initialize_client() -> OpenSearch:
         credentials = boto3.Session().get_credentials()
         if credentials:
             aws_auth = AWS4Auth(
-                credentials.access_key,
-                credentials.secret_key,
-                aws_region,
-                OPENSEARCH_SERVICE,
-                session_token=credentials.token
+                refreshable_credentials = credentials,
```
nathaliellenaa: Nit: remove spaces around `=` to follow the Python style guideline ([reference](https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements:~:text=Don%E2%80%99t%20use%20spaces%20around%20the%20%3D%20sign%20when%20used%20to%20indicate%20a%20keyword%20argument%2C%20or%20when%20used%20to%20indicate%20a%20default%20value%20for%20an%20unannotated%20function%20parameter%3A))

nathaliellenaa:
Add changelog?

src/opensearch/client.py
rithinpullela: Thanks Nathalie, will change it

nathaliellenaa: Sorry my comment was not really clear earlier, what I mean is to remove the spaces around `=` for keyword argument only, so we can keep `credentials = boto3.Session().get_credentials()` and `client_kwargs['http_auth'] = aws_auth` as is, and remove spaces for keyword arguments inside `AWS4Auth`

rithinpullela: Ahh got it, will change it.
We also need to work on adding linters etc to the repo to avoid such mistakes.

PR #20, opened 2025-06-04
Added URL as an optional parameter for all the tools

### LINKED ISSUE #15: [FEATURE] Support dynamic opensearch endpoints
_Currently, OpenSearch server endpoint needs to be provided as an environment variable. This restricts agents to make use of single opensearch endpoint rather than allowing agents to dynamically specifying endpoints._

_OpenSearch MCP server should optionally accept an endpoint in all supported tools. Provided endpoint should be used for OpenSearch API invocation. If an endpoint is not specified, then the default endpoint provided in environment variable needs to be used._

_OpenSearch endpoint URL could be passed as a url parameter._

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

rithin-pullela-aws:
### Description
- OpenSearch URL can not be accepted as an parameter for every tool. 
- If the URL is not provided by the MCP client, we fall back to the URL set via environment variable. 
- If the URL is not set via env variable and not passed in the tool call, we throw an error.

### Issues Resolved
Closes #15 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 
 ### Added
-
+- Add OpenSearch URl as an optional parameter for tool calls ([20](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/20))
 ### Removed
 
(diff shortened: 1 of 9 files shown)


nathaliellenaa (approved):
LGTM

Issue #7, opened 2025-05-23
[FEATURE] Expose tools based on OpenSearch Version


rithinpullela:
### Is your feature request related to a problem?
Yes. As OpenSearch evolves, newer tools (e.g., for creating connectors) are introduced that are not compatible with older versions. This can lead to unnecessary confusion or hallucinations in language models when they suggest unsupported tools for a given OpenSearch version.
### What solution would you like?
I would like to enforce version-based tool exposure. Specifically, when tools are created, each should have an associated minimum supported OpenSearch version. The system should detect the current OpenSearch version and only expose tools that are compatible with it.

### What alternatives have you considered?
An alternative could be manually maintaining tool availability by version or relying on documentation/user awareness. However, this approach is error-prone and does not scale well as more tools and versions are introduced.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

rithin-pullela-aws:
Closing the issue, resolved by #32 
Thank you @ssunno!!

PR #32, opened 2025-06-11
Feat: expose tool by version

ssunno:
### Description
This PR implements the tool filtering feature: When initializing the server, it retrieves the OpenSearch version and compares it with the versions specified in the TOOL_REGISTRY. Then, it filters the tools to include only those compatible with the OS version.

Set min_version or max_version in the TOOL_REGISTRY. The default is "no restriction."

### Issues Resolved
This PR will solve #7

### DIFF
--- a/src/tools/tools.py
+++ b/src/tools/tools.py
@@ -7,79 +7,75 @@ from typing import Any
 import json
 import os
 
+
 class ListIndicesArgs(BaseModel):
     opensearch_url: str = os.getenv("OPENSEARCH_URL", "")
+
 
 class GetIndexMappingArgs(BaseModel):
     index: str
     opensearch_url: str = os.getenv("OPENSEARCH_URL", "")
+
 
 class SearchIndexArgs(BaseModel):
     index: str
     query: Any
     opensearch_url: str = os.getenv("OPENSEARCH_URL", "")
 
+
 class GetShardsArgs(BaseModel):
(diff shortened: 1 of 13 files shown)


rithin-pullela-aws:
> But In my opinion, it is better to use decorator than TOOL_REGISTRY.
If the number of tools increases or they need more properties, managing them in the TOOL_REGISTRY dictionary may become harder.

The decorator is available by default in FastMCP server. We can build our own decorators or move to Fast MCP from the low level server for the support. But I agree that we need to create a simpler way to define tool properties

src/tools/tools.py
```diff
@@ -91,22 +87,19 @@ async def get_shards_tool(args: GetShardsArgs) -> list[dict]:
...
+
 
 TOOL_REGISTRY = {
     "ListIndexTool": {
         "description": "Lists all indices in OpenSearch",
         "input_schema": ListIndicesArgs.model_json_schema(),
         "function": list_indices_tool,
         "args_model": ListIndicesArgs,
+        "min_version": "1.0.0",
+        "max_version": "2.20",
```
rithinpullela: The List Index tool is supported in later versions too. So max version does not make sense here

ssunno: Sorry I forgot to remove this property after I tested it.

ssunno:
changelog added

PR #33, opened 2025-06-11
Add more tools from OpenSearch API specification

nathaliellenaa:
### Description
Add more tools from OpenSearch API specification
- `ClusterHealthTool` : Returns basic information about the health of the cluster.
- `CountTool` : Returns number of documents matching a query.
- `MsearchTool` : Allows to execute several search operations in one request.
- `ExplainTool` : Returns information about why a specific document matches (or doesn't match) a query.

These tools (performing write operations) will be added in future release along with the tool filtering feature:
- `IndicesCreateTool` : Creates an index with optional settings and mappings.
- `IndicesPutMappingTool` : Updates the index mappings.

### Issues Resolved


README.md
```diff
@@ -30,6 +30,12 @@ pip install opensearch-mcp-server-py
 - IndexMappingTool: Retrieves index mapping and setting information for an index in OpenSearch.
 - SearchIndexTool: Searches an index using a query written in query domain-specific language (DSL) in OpenSearch.
 - GetShardsTool: Gets information about shards in OpenSearch.
+- ClusterHealthTool : Returns basic information about the health of the cluster.
+- IndicesCreateTool : Creates an index with optional settings and mappings.
+- IndicesPutMappingTool : Updates the index mappings.
+- CountTool : Returns number of documents matching a query.
+- MsearchTool : Allows to execute several search operations in one request.
+- ExplainTool : Returns information about why a specific matches (or doesn't match) a query.
```
rithinpullela: nit: do you mean why a specific hit^ matches?

nathaliellenaa: Good catch! It should be `a specific document matches` ([ref](https://docs.opensearch.org/docs/latest/api-reference/search-apis/explain/)), the description is auto generated from the [API spec](https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/namespaces/_core.yaml#L1488). I'll update the readme and override the `ExplainTool` description manually with the correct description.

src/mcp_server_opensearch/stdio_server.py
```diff
@@ -9,16 +9,26 @@
...
+    # Call tool generator
+    await generate_tools_from_openapi(initialize_client(opensearch_url))
+
+    # Filter all tools by version
     version = get_opensearch_version(opensearch_url)
     enabled_tools = get_enabled_tools(version)
     logging.info(f"Connected OpenSearch version: {version}")
     logging.info(f"Enabled tools: {list(enabled_tools.keys())}")
 
+    await generate_tools_from_openapi(initialize_client(opensearch_url))
```
rithinpullela: Is there a reason why we have this line in 2 places? 
Line 22 and 30

nathaliellenaa: Good catch, line 30 should be deleted

README.md
```diff
@@ -30,6 +30,12 @@ pip install opensearch-mcp-server-py
 - IndexMappingTool: Retrieves index mapping and setting information for an index in OpenSearch.
 - SearchIndexTool: Searches an index using a query written in query domain-specific language (DSL) in OpenSearch.
 - GetShardsTool: Gets information about shards in OpenSearch.
+- ClusterHealthTool : Returns basic information about the health of the cluster.
+- IndicesCreateTool : Creates an index with optional settings and mappings.
+- IndicesPutMappingTool : Updates the index mappings.
```
rithinpullela: Just thinking out loud, IndicesPutMappingTool and IndicesCreateTool perform write operations, should we expose them after we have a Tool Filtering option through which the users can allow and deny the tools during the setup

nathaliellenaa: Hmm I've also thought about this before, but these 2 tools are requested by the music team. If they're ok with pushing these two tools into the subsequent release (not the upcoming one), we can include them in the subsequent release along with the tool filtering feature.

rithinpullela: I agree.

src/mcp_server_opensearch/stdio_server.py
```diff
@@ -9,16 +9,26 @@
 import os
 import logging
 
+from tools.tools import TOOL_REGISTRY
```
rithinpullela: Looks like we are not using this variable. Maybe remove this?

src/mcp_server_opensearch/sse_server.py
```diff
@@ -17,10 +17,18 @@
 import os
 import logging
 
+from tools.tools import TOOL_REGISTRY
```
rithinpullela: Looks like we are not using this variable. Maybe remove this?

rithin-pullela-aws (approved):

rithin-pullela-aws:
Thanks Nathalie, approved the PR

PR #35, opened 2025-06-16
Add support for Multiple OpenSearch cluster Connectivity

rithin-pullela-aws:
### Description
With these changes, the MCP server can not be instantiated in 2 modes with --mode flag.

**Single Mode**: Same behavior as the current MCP server, i.e pick up credentials from environment variables
**Multi Mode**: The MCP Server accepts extra variables for connecting to different clients

Muti mode allows `opensearch_url`, `iam_arn`, `aws_region`, `opensearch_username`, and `opensearch_password` as optional parameters.

### Design:
Multi-mcp

### Setup expected from customers:

**Start  the server in Multi mode:**

* Start the server with --mode as multi

**IAM Roles setup (if needed):**

* Create an IAM role for each cluster.
* For all these IAM roles add Trust relationship to be assumed by a user or role

Example:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "Statement1",
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::419213735998:root" // I used a root user in this example
            },
            "Action": "sts:AssumeRole"
        }
    ]
}
```
**Env setup:**
* The MCP server tries to assume the IAM role provided by the LLM during tool call.
* We need to add the credentials of the role/user which has permissions to assume all the IAM roles in environment variables (or via aws configure or if running on EC2 instance, instance profile role)
* These credentials will be used by the MCP server to assume IAM roles.
* If only basic auth is used, we do not need to do this setup.

**Provide Context to the LLM:**

The LLM to successfully make calls to different clusters, the users need to provide context about the connection information. This can be done my manually putting the information in the prompt.

### Issues Resolved

### DIFF
--- a/src/opensearch/client.py
+++ b/src/opensearch/client.py
@@ -8,6 +8,7 @@ import os
 import boto3
 import logging
 from typing import Dict, Any
+from .tool_params import baseToolArgs
 
 # Configure logging
 logging.basicConfig(level=logging.INFO)
@@ -18,7 +19,7 @@ OPENSEARCH_SERVICE = "es"
 OPENSEARCH_SERVERLESS_SERVICE = "aoss"
 
 # This file should expose the OpenSearch py client
-def initialize_client(opensearch_url: str) -> OpenSearch:
+def initialize_client(args: baseToolArgs) -> OpenSearch:
     """
     Initialize and return an OpenSearch client with appropriate authentication.
     
@@ -29,7 +30,7 @@ def initialize_client(opensearch_url: str) -> OpenSearch:
        - Uses 'es' service name otherwise
 
     Args:
-        opensearch_url (str): The URL of the OpenSearch cluster. Must be a non-empty string.
+        args (baseToolArgs): The arguments object containing authentication and connection details
     
     Returns:
         OpenSearch: An initialized OpenSearch client instance.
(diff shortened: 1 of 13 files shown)


src/opensearch/client.py
```diff
@@ -52,33 +55,62 @@ def initialize_client(opensearch_url: str) -> OpenSearch:
...
-        'hosts': [opensearch_url],
+        'hosts': [args.opensearch_url],
         'use_ssl': (parsed_url.scheme == "https"),
         'verify_certs': True,
         'connection_class': RequestsHttpConnection,
     }
 
+    session = boto3.Session()
+    if not args.aws_region:
+        args.aws_region = os.getenv("AWS_REGION") or session.region_name
```
jiapingzeng, line 70: Could we parse region from `args.opensearch_url`? Would reduce the number of required arguments.

Note that the region position within URL is different for dual stack and IPv4 only:
dual stack endpoint: https://search-jpz-test-xxxxx.aos.us-east-1.on.aws
IPv4 endpoint: https://search-jpz-test-xxxxx.us-east-1.es.amazonaws.com

rithinpullela: That is a really good suggestion, can we be sure all the URLs have region?

nathaliellenaa: I think it's possible to create a custom endpoint for AOS, and users might not specify the region in their custom URL ([reference](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/customendpoint.html)). Maybe we can first check if there is region in the URL, if not it will fall back to the env variables.

rithinpullela: Great suggestion, makes sense let us add that. I will update readme and other files along with URL parsing to find region.

PR #40, opened 2025-06-21
add `OPENSEARCH_SSL_VERIFY` environment variable

jiapingzeng:
### Description
add `OPENSEARCH_SSL_VERIFY` environment variable to allow not verifying certificates

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-mcp-server-py/issues/38

### DIFF
--- a/src/opensearch/client.py
+++ b/src/opensearch/client.py
@@ -80,8 +80,8 @@ def initialize_client_with_cluster(cluster_info: ClusterInfo = None) -> OpenSear
     # Common client configuration
     client_kwargs: Dict[str, Any] = {
         'hosts': [opensearch_url],
-        'use_ssl': (parsed_url.scheme == "https"),
-        'verify_certs': os.getenv("OPENSEARCH_SSL_VERIFY", True),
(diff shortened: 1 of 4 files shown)


src/opensearch/client.py
```diff
@@ -58,7 +58,7 @@ def initialize_client(opensearch_url: str) -> OpenSearch:
     client_kwargs: Dict[str, Any] = {
         'hosts': [opensearch_url],
         'use_ssl': (parsed_url.scheme == "https"),
-        'verify_certs': True,
+        'verify_certs': os.getenv("SSL_VERIFY", True),
```
rithinpullela: Can we name it `OPENSEARCH_SSL_VERIFY`?

rithin-pullela-aws:
Hi @jiapingzeng, I changed the way we handle the env variable to `os.getenv('OPENSEARCH_SSL_VERIFY', 'true').lower() != 'false'`

If the env variable is set to a random string like `abc`, we want to have ssl verify as true. So unless we explicitly set it to false, it will be considered as true.

ylwu-amzn (approved):

Issue #42, opened 2025-06-21
[BUG] OpenSearch Server Less


hksharma-ai:
OpenSearch API call work fine with the postman but fail with claude desktop with 403  error code where as we  are passing same credential in both the cases.

hksharma-ai:
NFO:opensearch.client:Using AWS region: us-west-2
INFO:opensearch.client:Using OpenSearch Serverless with service name: is_serverless True
INFO:opensearch.client:Using OpenSearch Serverless with service name: aoss
INFO:opensearch.client:Using OpenSearch parsed_url: parsed_url ParseResult(scheme='https', netloc='xyz.us-west-2.aoss.amazonaws.com:443', path='/_cat', params='', query='', fragment='')


INFO:opensearch.client:opensearch_url is: https://xyz.us-west-2.aoss.amazonaws.com:443/_cat


INFO:opensearch.client:print client_kwargs : {'hosts': ['https://xyz.us-west-2.aoss.amazonaws.com:443/_cat'], 'use_ssl': True, 'verify_certs': True, 'connection_class': <class 'opensearchpy.connection.http_requests.RequestsHttpConnection'>}
INFO:botocore.credentials:Found credentials in environment variables.
INFO:opensearch.client:Using AWS IAM authentication
INFO:opensearch.client:Using AWS region: us-west-2




WARNING:opensearch:GET https://xyz.us-west-2.aoss.amazonaws.com:443/_cat/ [status:404 request:3.725s]
Traceback (most recent call last):

arjunkumargiri:
Is url set as `[https://xyz.us-west-2.aoss.amazonaws.com:443/_cat](https://xyz.us-west-2.aoss.amazonaws.com/_cat)` can you try with [https://xyz.us-west-2.aoss.amazonaws.com](https://xyz.us-west-2.aoss.amazonaws.com)

hksharma-ai:
I tried that too. What I observed that serverless OS never exposed any health endpoint on base URL
https://xyz.us-west-2.aoss.amazonaws.com/

hksharma-ai:
response= client.info();
print("\nCreating Info:")

This will always throw the 

NotFoundError(404, '')

hksharma-ai:
As per my research

This is a known behavior with Amazon OpenSearch Serverless:
[...]

Issue #29, opened 2025-06-10
[FEATURE] Support IAM Profile Auth


mateon01:
### Is your feature request related to a problem?
Customers should be able to access OpenSearch through a profile specified separately from the default.

### What solution would you like?
It must be able to receive AWS_PROFILE as args along with OpenSearch Url.

### What alternatives have you considered?
Currently, only the Key is accepted, and if it is not available, the Default profile is used. Some customers use different profiles for multiple accounts.

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._

rithin-pullela-aws:
This has been addressed in #45 and released in the 0.2 version. Closing this task.

Issue #26, opened 2025-06-06
[FEATURE] Support IAM role based authN


apuljain:
### Is your feature request related to a problem?
Currently we have to specify AWS_SECRET_ACCESS_KEY as env variable for the MCP server to work. This is not safe from security perspective. We would like to have IAM role support instead where client could specify IAM role ARN as the env variable and MCP server could use it to access OS domain.

### What solution would you like?
_A clear and concise description of what you want to happen._
We would like to have IAM role support instead where client could specify IAM role ARN as the env variable and MCP server could use it to access OS domain

### What alternatives have you considered?
_A clear and concise description of any alternative solutions or features you've considered._
N/A

### Do you have any additional context?
_Add any other context or screenshots about the feature request here._
N/A

rithin-pullela-aws:
Hi @apuljain, thanks for creating the issue. Currently we do not need to explicitly specify `AWS_SECRET_ACCESS_KEY` as an env variable. 

OpenSearch MCP Server using `boto3` to get the credentials.
When you call `boto3.Session().get_credentials(),` Boto3 searches for AWS credentials using a specific order known as the credential provider chain. It stops as soon as it finds valid credentials. Here's the order Boto3 follows:
- **Explicitly passed credentials** – Credentials provided directly when creating the `boto3.client()` or `boto3.Session() `object.
- **Environment variables** – Boto3 checks for `AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_SESSION_TOKEN`.
- **Assume Role provider** – Uses roles specified in the AWS config file to obtain temporary credentials.
- **Assume Role with Web Identity provider** – Uses web identity tokens to assume a role.
- **AWS IAM Identity Center (SSO) provider** – Retrieves credentials through AWS IAM Identity Center if configured.
- **Shared credentials file** – Located at `~/.aws/credentials`.
- **AWS config file** – Located at `~/.aws/config`.
- **Boto2 config file** – Checks legacy config files at `/etc/boto.cfg` and `~/.boto`.
- **Container credentials** – Used if running in a container environment like ECS.
- **Instance metadata service** – If running on an EC2 instance with an IAM role, credentials are pulled from the instance metadata.

Reference: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html?utm_source=chatgpt.com#configuring-credentials


[...]

Issue #25, opened 2025-06-05
[FEATURE]Support Multiple Opensearch Domains


cgunawat:
Enable the Model Context Protocol (MCP) configuration to support connections to multiple OpenSearch domains to allow for broader context ingestion across environments and use cases.

Problem Statement:
Currently, MCP supports only a single OpenSearch domain in its configuration. This limitation restricts the ability to query and correlate contextual data that may be distributed across different OpenSearch domains — for example, logs in one domain, crash data in another, and metrics in a third.

This would enable:

Aggregated or federated querying across domains

rithin-pullela-aws:
Thanks for creating the issue @cgunawat, I think this will be a good use case empowering the agents to perform better.

rithin-pullela-aws:
This feature is supported via #45, hence closing the issue

PR #45, opened 2025-06-24
Add support for Multiple OpenSearch cluster Connectivity

rithinpullela:
### Description
To connect to multiple clusters, instead of getting all the information from the LLM. We accept a file path as a parameter called `--config`. We expect a yml file in this path which has information about different clusters.

The format of the yml file:
```
version: "1.0"
description: "OpenSearch cluster configurations"


clusters:
  local-cluster:
    opensearch_url: "http://localhost:9200"
    opensearch_username: "admin"
    opensearch_password: "your_password_here"

  remote-cluster:
    opensearch_url: "https://your-opensearch-domain.us-east-2.es.amazonaws.com"
    profile: "your-aws-profile"

  remote-cluster-with-iam:
    opensearch_url: "https://your-opensearch-domain.us-east-2.es.amazonaws.com"
    iam_arn: "arn:aws:iam::123456789012:role/YourOpenSearchRole"
    aws_region: "us-east-2"
    profile: "your-aws-profile"


# Example configurations for different authentication methods:
# 
# 1. IAM Role Authentication (recommended for production):
#    - Requires: opensearch_url, iam_arn, aws_region, profile
#    - Uses AWS IAM roles for authentication
#
# 2. Basic Authentication:
#    - Requires: opensearch_url, opensearch_username, opensearch_password
#    - Uses username/password for authentication
#
# 3. AWS Profile Authentication:
#    - Requires: opensearch_url, profile
#    - Uses AWS credentials from the specified profile

```

The config file accepts basic auth, IAM auth, and direct connection using AWS credentials. If you have different credentials in different aws profiles, the users can add profile parameter too. The MCP server will use creds from the profile mentioned.

**Single vs Multi mode:**
- To keep the interface simple and intuitive, we separate the single and multi modes with the flag `--mode` defaulting to single mode.
- The tools multi mode accepts another parameter, the `opensearch_cluster_name`. This parameter is provided by the LLM. We get the connection information for the corresponding cluster name from the config file.
- The single mode supports tool filtering based on version, while the multi mode raises an error in the tool call when there is a version mismatch.
- If config file path is not provided, the multi mode falls back to single mode behavior.

**--profile command line argument**
- We also accept the --profile variable which connects to clusters using aws credentials from that particular profile.
- The profile variable is also accepted in config file.

### Issues Resolved
#17, #25 , #26, #29, #42

### DIFF
--- a/tests/mcp_server_opensearch/test_clusters_information.py
+++ b/tests/mcp_server_opensearch/test_clusters_information.py
@@ -2,6 +2,7 @@ import pytest
 import tempfile
 import os
 import yaml
+import platform
 from unittest.mock import patch, MagicMock
 from mcp_server_opensearch.clusters_information import (
     ClusterInfo,
@@ -206,6 +207,10 @@ clusters:
 
     def test_load_clusters_from_yaml_permission_error(self):
         """Test loading from file with permission error."""
+        # Skip this test on Windows as os.chmod with 0o000 doesn't work reliably
+        if platform.system() == 'Windows':
+            pytest.skip("Permission test not reliable on Windows")
+            
         with tempfile.NamedTemporaryFile(mode='w', suffix='.yml', delete=False) as f:
             f.write('clusters: {}')
             f.flush()


rithin-pullela-aws:
Will update user guide, change log, and add more UTs.

hksharma-ai:
you still initialise the client multiple times. why you need to do that ?. Make client initialisation global and use that client object wherever required

nathaliellenaa:
> We accept a file path as a parameter called --clusters_config

We'll have something similar like this for tool filter through YAML file. Do you think it'll be better to make the name more general, for example `--config <path to a config yaml file>`? The YAML file can include both the multi cluster and tool filter setup.

src/mcp_server_opensearch/__init__.py
```diff
@@ -4,40 +4,56 @@
...
-        default='0.0.0.0', 
-        help='Host to bind to (SSE only)'
+        '--profile', default='', help='AWS profile to use for OpenSearch connection'
     )
     parser.add_argument(
-        '--port', 
-        type=int, 
-        default=9900, 
-        help='Port to listen on (SSE only)'
+        '--clusters_config', default='', help='YAML file containing cluster information'
```
nathaliellenaa: We'll have something similar like this for tool filter through YAML file. Do you think it'll be better to make the name more general, for example `--config <path to a config yaml file>`? The YAML file can include both the multi cluster and tool filter setup.

src/mcp_server_opensearch/clusters_information.py
```diff
@@ -0,0 +1,152 @@
...
+
+def load_clusters_from_yaml(file_path: str) -> None:
+    """Load cluster configurations from a YAML file and populate the global registry.
+
+    Args:
+        file_path: Path to the YAML configuration file
+
+    Raises:
+        FileNotFoundError: If the YAML file doesn't exist
+        yaml.YAMLError: If the YAML file is malformed
```
nathaliellenaa, line 59: Better to also check for `PermissionError` when reading the file from the given path

src/mcp_server_opensearch/clusters_information.py
```diff
@@ -0,0 +1,152 @@
...
+    result = {'loaded_clusters': [], 'errors': [], 'total_clusters': 0}
+
+    try:
+        with open(file_path, 'r', encoding='utf-8') as file:
+            config = yaml.safe_load(file)
+
+        # Process clusters
+        clusters = config.get('clusters', {})
+        result['total_clusters'] = len(clusters)
+        logging.info(f'total cluster found in config file: {result["total_clusters"]}')
```
nathaliellenaa:
```suggestion
        logging.info(f'Total cluster found in config file: {result["total_clusters"]}')
```

src/opensearch/client.py
```diff
@@ -1,79 +1,131 @@
...
+        raise ValueError(
+            'OpenSearch URL must be provided using config file or OPENSEARCH_URL environment variable'
+        )
+    opensearch_username = (
+        cluster_info.opensearch_username if cluster_info else os.getenv('OPENSEARCH_USERNAME', '')
+    )
+    opensearch_password = (
+        cluster_info.opensearch_password if cluster_info else os.getenv('OPENSEARCH_PASSWORD', '')
+    )
+    aws_region = cluster_info.aws_region if cluster_info else ''
```
nathaliellenaa, line 64: Should we add a fallback check for `AWS_REGION` in the env variable?

src/mcp_server_opensearch/__init__.py
rithinpullela: Makes sense, will do that thank you!

src/opensearch/client.py
rithinpullela: Hi @nathaliellenaa , this was intentional. In line 89-90, I am reading from AWS_REGION variable.

The priority for picking up aws_region is: cluster config > session > env variable.

This is done keeping in mind the multi cluster case

rithin-pullela-aws:
> you still initialise the client multiple times. why you need to do that ?. Make client initialisation global and use that client object wherever required

Absolutely agreed that we need to cache the instance of the client. The reason why I am initializing for every request is to support the SSE case. In streaming/ sse, we will have multiple clients connecting to the server. It might make sense to cache client based on session ID in that case. Will need a larger effort and discussion to deal with. I will start a conversation in the issues soon.

DEVELOPER_GUIDE.md
```diff
@@ -24,77 +35,150 @@ source .venv/bin/activate
...
-## Sample MCP config file for development:
+### Multi Mode Development
+```bash
+cd src
+
+# Run with custom config file
+uv run python -m mcp_server_opensearch --mode multi --config ../config/dev-clusters.yml
+
+# Run SSE server with multi mode
+uv run python -m mcp_server_opensearch --mode multi --config ../config/dev-clusters.yml --transport sse
```
nathaliellenaa:
```suggestion
# Run stdio server in multi mode with custom config file
uv run python -m mcp_server_opensearch --mode multi --config ../config/dev-clusters.yml

# Run SSE server in multi mode with custom config file
uv run python -m mcp_server_opensearch --mode multi --config ../config/dev-clusters.yml --transport sse
```

DEVELOPER_GUIDE.md
```diff
@@ -24,77 +35,150 @@ source .venv/bin/activate
...
 ```
+
+## Development Configuration
+
+### MCP Configuration for Development
+
+Create an MCP configuration file for your AI agent to connect to your development server:
+
+#### For Q Developer CLI (`~/.aws/amazonq/mcp.json`):
```
nathaliellenaa, line 86: Seems like the config file for Q CLI and Claude Desktop is the same, can we combine them into one to make the guide more concise?

DEVELOPER_GUIDE.md
```diff
@@ -24,77 +35,150 @@ source .venv/bin/activate
...
                 "--",
                 "python",
                 "-m",
                 "mcp_server_opensearch"
             ],
             "env": {
-                // Optional
                 "OPENSEARCH_URL": "<your_opensearch_domain_url>",
+                "OPENSEARCH_USERNAME": "<your_username>",
+                "OPENSEARCH_PASSWORD": "<your_password>"
```
nathaliellenaa:
```suggestion
                "OPENSEARCH_USERNAME": "<your_opensearch_domain_username>",
                "OPENSEARCH_PASSWORD": "<your_opensearch_domain_password>"
```

USER_GUIDE.md
```diff
@@ -1,117 +1,221 @@
...
+    profile: "staging"
 ```
 
-### Sample configuration
+**Key Points about Multi Mode:**
+- **Cluster Names**: The keys under `clusters` (e.g., `local-dev`, `production`, `staging`) are used as the `opensearch_cluster_name` parameter when calling tools
+- **Authentication**: Each cluster can use different authentication methods (basic auth, IAM roles, AWS profiles)
+- **Tool Usage**: When using tools, you must specify which cluster to use: `{"opensearch_cluster_name": "production", "index": "users"}`
+
+See [Authentication](#authentication) sections for detailed setup instructions.
```
nathaliellenaa: Seems like this is also mentioned in line 132, maybe remove this line to avoid duplication?

example_clusters.yml
```diff
@@ -0,0 +1,34 @@
+version: "1.0"
```
nathaliellenaa, line 1: Just curious, is there any reason why we need to have `version` in the config file?

src/mcp_server_opensearch/clusters_information.py
```diff
@@ -0,0 +1,165 @@
...
+def list_clusters() -> list[str]:
+    """Get list of all registered cluster names.
+
+    Returns:
+        list[str]: List of cluster names
+    """
+    return list(cluster_registry.keys())
+
+
+def get_all_clusters() -> Dict[str, ClusterInfo]:
```
nathaliellenaa: I don't see `list_clusters` and `get_all_clusters` methods being used, should we remove them?

rithinpullela: Yes makes sense. Was planning on using these functions, but did not need to. Will clean it up

example_clusters.yml
rithinpullela: Just a good practice, in the future if we have different versions of the config file which we want to parse in a different way, this would be helpful

PR #46, opened 2025-06-25
Add tool filter feature

nathaliellenaa:
### Description
Add both approaches (YAML config file and environment variables) of tool filter feature. Note that the tool filter feature is now supported in the `single` mode only, meaning that we don't support tool filtering in multi cluster connectivity. More details about the user experience is written in the [user guide](https://github.com/opensearch-project/opensearch-mcp-server-py/blob/c92aab24ec341257deebd4974af463077a41416f/USER_GUIDE.md).

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-mcp-server-py/issues/37


USER_GUIDE.md
```diff
@@ -316,6 +317,84 @@ python -m mcp_server_opensearch --mode multi --config clusters.yml --profile my-
...
+
+# Run SSE server with tool filter config file
+python -m mcp_server_opensearch --transport sse --config path/to/config.yml
+```
+
+2. Environment Variables
+Set environment variables for tool filtering:
+```bash
+# Tool Categories
+export OPENSEARCH_TOOL_CATEGORIES='{"<name_of_category>":["<tool_name>","<tool_name>"]}'
```
rithinpullela, line 362: If we want to create multiple of these categories, we just need to add more categories in the json right?

USER_GUIDE.md
```diff
@@ -316,6 +317,84 @@ python -m mcp_server_opensearch --mode multi --config clusters.yml --profile my-
...
+# Operation Settings
+export OPENSEARCH_SETTINGS_ALLOW_READ=true
+export OPENSEARCH_SETTINGS_ALLOW_WRITE=false
+```
+
+### Important Notes
+- Tool names are case-insensitive
+- All configuration fields are optional
+- Priority order (highest to lowest):
+    - `enable_tools`/`disable_tools`
```
rithinpullela: If a tool is enabled and disabled, which one takes the priority? 

Update: I can see from the code, if a tool is listed in any kind of enable and any kind of disable we are prioritizing the disable. Can we update to read me to reflect the same too?

pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "mcp[cli]>=1.6.0",
     "opensearch-py>=2.8.0",
     "pydantic>=2.11.3",
+    "pytest-cov>=6.2.1",
```
rithinpullela: nit: Can we add this to dev dependencies

src/tools/tool_filter.py
```diff
@@ -1,15 +1,210 @@
...
 import logging
 from .tool_params import baseToolArgs
 from .tools import TOOL_REGISTRY
 from .utils import is_tool_compatible
 from opensearch.helper import get_opensearch_version
-import os
 
 
-def get_tools(mode: str = 'single') -> dict:
+def parse_comma_separated(text, separator=","):
```
rithinpullela: nit: maybe we can have this func in utils?

src/tools/tool_filter.py
```diff
@@ -31,6 +226,29 @@ def get_tools(mode: str = 'single') -> dict:
...
+    }
+
+    # Check if both config and env variables are set
+    if config and any(env_config.values()):
+        logging.warning("Both config file and environment variables are set. Using config file.")
+
+    # Apply tool filtering
+    process_tool_filter(
+        filter_path=config if config else None,
+        **{k: v for k, v in env_config.items() if config is ''}
```
rithinpullela: In the second parameter, can we use `not config` here? It checks for both none and empty type.

src/tools/tool_filter.py
```diff
@@ -1,15 +1,210 @@
...
+def process_tool_filter(
+        enabled_tools: str = None,
+        disabled_tools: str = None,
+        tool_categories: str = None,
+        enabled_categories: str = None,
+        disabled_categories: str = None,
+        enabled_tools_regex: str = None,
+        disabled_tools_regex: str = None,
+        allow_read: bool = None,
+        allow_write: bool = None,
```
rithinpullela: Can we have a single variable to track allow read and allow right? I am assuming our aim to expose an easy way to filter out write tools. So we can have a single variable called allow_write defaulted to true. I don't see a point in disabling the read tools.

src/tools/tool_filter.py
```diff
@@ -1,15 +1,210 @@
...
+            enabled_category_list = tool_filters.get('enabled_categories', [])
+            disabled_category_list = tool_filters.get('disabled_categories', [])
+            enabled_tools_regex_list = tool_filters.get('enabled_tools_regex', [])
+            disabled_tools_regex_list = tool_filters.get('disabled_tools_regex', [])
+            
+            # Get settings
+            settings = tool_filters.get('settings', {})
+            if settings:
+                allow_read = settings.get('allow_read', allow_read)
+                allow_write = settings.get('allow_write', allow_write)
```
rithinpullela: Are we defaulting to environment variable here? It could be confusing for the users if they forget to unset their env variables. Maybe when they give the config file, we should only consider the config file and use default values if not provided. i.e both true?

src/tools/tool_filter.py
```diff
@@ -1,15 +1,210 @@
...
+
+        # Process environment variables
+        if tool_categories:
+            try:
+                category_to_tools.update(json.loads(tool_categories) if isinstance(tool_categories, str) else {})
+            except json.JSONDecodeError:
+                logging.warning(f"Invalid JSON in tool_categories: {tool_categories}")
+        
+        # Parse comma-separated strings from environment variables
+        if enabled_tools:
```
rithinpullela: Here we are overriding the config file setup with env variables, is that the expected priority order? env > config?

USER_GUIDE.md
```diff
@@ -316,6 +317,84 @@ python -m mcp_server_opensearch --mode multi --config clusters.yml --profile my-
...
+
+# Operation Settings
+export OPENSEARCH_SETTINGS_ALLOW_READ=true
+export OPENSEARCH_SETTINGS_ALLOW_WRITE=false
+```
+
+### Important Notes
+- Tool names are case-insensitive
+- All configuration fields are optional
+- Priority order (highest to lowest):
```
rithinpullela: In the code, I saw we are enabling all the tools mentioned in any of the enable type followed by disabling the tools in any of the disable category. So we have no priority among types like `enable_tools > enable_categories > enabled_tools_regex` right? Can you explain how this priority order works?

USER_GUIDE.md
nathaliellenaa: Yes, users can create multiple categories. For example `'{"category1":["ListIndexTool"],"category2":["SearchIndexTool"]}'`

USER_GUIDE.md
nathaliellenaa: Yes, the disable will be prioritized. Will update the doc

src/tools/tool_filter.py
nathaliellenaa: Sure, will change it

src/tools/tool_filter.py
nathaliellenaa: Makes sense, I'll remove the `allow_read`

src/tools/tool_filter.py
nathaliellenaa: Hmm makes sense, the `allow_read` will be removed and I'll add a default value for the `allow_write` to true.

src/tools/tool_filter.py
nathaliellenaa: The priority order should be config > env. If we see here, we check if both the config file or any of the env variables are set. If the config file is provided, it will choose the config file over the env variables https://github.com/opensearch-project/opensearch-mcp-server-py/blob/c92aab24ec341257deebd4974af463077a41416f/src/tools/tool_filter.py#L246-L250

USER_GUIDE.md
nathaliellenaa: Yes, we don't have priority among `enable_tools > enable_categories > enabled_tools_regex`. For example, users have a config file like this:
```
tool_category:
    critical:
    - ListIndexTool
    - SearchIndexTool
  
tool_filters:
    enabled_categories:
     - critical
    disabled_tools:
     - SearchIndexTool
```
We can see here that users enable `critical` category that has `ListIndexTool` and `SearchIndexTool`. However, since `disabled_tools` has higher priority compare to `enable_categories`, the `SearchIndexTool` will be disabled and at the end, users will only see the `ListIndexTool`. I hope this makes sense

src/tools/tool_filter.py
rithinpullela: Ahh, makes sense. Thanks for clarifying Nathalie!

USER_GUIDE.md
rithinpullela: Thanks Nathalie, and I agree with the behavior of the feature. But the User guide might be suggesting otherwise: 

```
Priority order (highest to lowest):
enable_tools/disable_tools
enable_categories/disable_categories
enabled_tools_regex/disabled_tools_regex
```
Sounds like we prioritize entries in `enable_tools/disable_tools` over entries in `enable_categories/disable_categories`

nathaliellenaa: Yes, the user guide suggests that `enable_tools/disable_tools` will be prioritized over `enable_categories/disable_categories`. This reflects the behavior I mentioned before where `disabled_tools` has higher priority compare to `enable_categories`.

src/tools/utils.py
```diff
@@ -21,3 +24,67 @@ def is_tool_compatible(current_version: Version, tool_info: dict = {}):
...
+    for tool in tool_list:
+        tool_lower = tool.lower()
+        if tool_lower in registry_lookup:
+            valid_tools.add(tool_lower)
+        else:
+            logging.warning(f"Ignoring invalid tool from '{source_name}': '{tool}'")
+    return valid_tools
+
+
+def apply_write_filter(registry, allow_write):
```
rithinpullela: Looks like we are not using allow_write variable, maybe remove that?

Also, nit: Generally we put the functions that are generic / reusable / not core of the business logic in the utils file, for example string processing functions etc. Since the functions `parse_comma_separated`, `load_yaml_config ` are generic and could be potentially used in other files, it makes sense for them to stay in utils, but the functions `process_regex_patterns`, `apply_write_filter `, and `process_categories ` are only used in a single file and quite specific to that use case. Maybe we can have these 3 functions in the tool_filter file itself?

nathaliellenaa: Got it, thanks for the explanation Rithin!

rithin-pullela-aws (approved):
LGTM! Thanks @nathaliellenaa

PR #68, opened 2025-07-02
Extend get indices tool

### LINKED ISSUE #66: [PROPOSAL] Return more details from the cat indices API
### What are you proposing?
The `list_indices` helper returns json from the cat api but the `list_indices_tool` drops all the extra data and converts the indices to a newline delimited list. This proposal is to make the other data available from the `list_indices` helper. 

https://github.com/opensearch-project/opensearch-mcp-server-py/blob/main/src/opensearch/helper.py#L11-L16

### What users have asked for this feature?
me

### What problems are you trying to solve?
I'd like to make the `list_indices_tool` more useful by including the other data (docs.count, docs.deleted, storage.size, pri.store.size). This opens the ability for users to use sort indices by size and begin asking more complex questions of their data in the cluster. 

### What is the developer experience going to be?
The experience should be the same.

#### Are there any security considerations? 
As far as I can tell this would not change the current security posture. It does expose a few extra details about the indices. See https://docs.opensearch.org/docs/latest/api-reference/cat/cat-indices/

#### Are there any breaking changes to the API
Depends. I can think of 2 ways to implement this and my preference is for the second method.

1. Alter `list_indices_tool` to return the full json from `list_indcies` helper. This is not backwards compatible and alters the user experience.
2. Add a secondary `list_indices_full_tool` to return the full json from `list_indices` helper. This would be a backwards compatible change that extends the mcp with a new tool.

### What is the user experience going to be?
Depends on implementation (see above). The users will have the ability to work with more data about the indices.

#### Are there breaking changes to the User Experience?
Depends on implementation (see above). 

### Why should it be built? Any reason not to?
I think this opens up a few more use cases for this mcp. 

As an example, investigating space hogs for a cluster and viewing the growth over time of rolling indices. AI agents can use the list_indices_tool (or list_indices_full_tool) to gather indices and their storage size then maybe a get_index_tool to gather the creation date.

### What will it take to execute?
I'd like to submit a PR to add 2 new tools and a new helper.

`list_indices_full_tool` - mimmic `list_indices_tool` but return the full json
`get_index_tool` - return [details](https://docs.opensearch.org/docs/latest/api-reference/index-apis/get-index/) for a single index from `get_index` helper 
`get_index` - a helper that uses get index api to gather settings for a single index

### Any remaining open questions?
The get_index_tool and get_index helper are extra enhancements that can be considered out-of-scope or TBD in the future.

renegaderyu:
### Description
- Extend get indices tool to return full response from cat api.
- Add an optional parameter to list_indices that if specified will return detailed data about the index from the get index api.

### Issues Resolved
This PR will resolve #66 


rithin-pullela-aws:
Thanks for raising the PR @renegaderyu! Can you please update the failing UTs?

renegaderyu:
> Thanks for raising the PR @renegaderyu! Can you please update the failing UTs?

Yeah, done.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fix endpoint selection bug in ClusterHealthTool and CountTool ([#59](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/59))
 - Fix Serverless issues ([#61](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/61))
+- Extend list indices tool ([#68](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/68))
```
nathaliellenaa: Nit: Can you move this line to the `Added` section on line 7?

nathaliellenaa (commented):
Thanks @renegaderyu, LGTM. Added a small nit comment

CHANGELOG.md
renegaderyu: 🤦 yeah totally wasn't thinking/paying attention there...fixed now.

Issue #72, opened 2025-07-15
[BUG] Using MCP with a limited read-only user


tomikonio:
Based on this discussion: https://github.com/opensearch-project/opensearch-mcp-server-py/discussions/70
### What is the bug?

Running the MCP server with a limited read-only basic auth user, getting this error:
```
AuthorizationException(403, 'security_exception', 'no permissions for [cluster:monitor/main] and User [name=bobi, backend_roles=[], requestedTenant=null]')
```

When trying to use  a URL that points to the specific index (i.e. `<URL>/<index_name>`), we get the follwoing error:

```
opensearchpy.exceptions.AuthorizationException: AuthorizationException(403, 'security_exception', 'no permissions for [indices:admin/get] and User [name=bobi, backend_roles=[], requestedTenant=null]')
```

When trying to run it with this path: `<URL>/<index_name>-*/_search`, we get the following error:
```
line 57, in get_opensearch_version
    return Version.parse(response['version']['number'])
                         ~~~~~~~~^^^^^^^^^^^
KeyError: 'version'
```

### How can one reproduce the bug?

1. Create an OpenSearch user with read-only access to a specific index.
2. Launch the MCP server using that user's basic-auth credentials.
3. Observe the error raised due to missing `cluster:monitor/main` permission.
4. Attempt to start the server with `<URL>/<index_name>` and get `AuthorizationException` for `[indices:admin/get]`
5.  Attempt to start the server with `<URL>/<index_name>-*/_search` and  get the `KeyError: 'version'`.

### What is the expected behavior?

The MCP server should support being run with limited-permission users by either:
- Avoiding use of operations that require broad cluster-level permissions (like `client.info()`), or
- Falling back gracefully when such operations fail, allowing the server to function in limited-permission contexts.

### What is your host/environment?

- **OpenSearch version:** 1.3.2  
- **Execution method:** MCP server started via `uvx opensearch-mcp-server-py`  
- **User privileges:** Access to a single index only (no cluster-level permissions), using basic auth.

### Do you have any additional context?

The root cause found by @rithin-pullela-aws: https://github.com/opensearch-project/opensearch-mcp-server-py/blob/51cd164055852971f24b7e02349cff5723b61ad4/src/opensearch/helper.py#L59C1-L73C56

#### Potential fixes proposed by @rithin-pullela-aws :
- Find a way to get version which works for all the users without relying on client.info()
- Accept an env variable to skip version based tool filtering
[...]

PR #74, opened 2025-07-19
Handle Tool Filtering failure gracefully and define priority to the AWS Region definitions

rithin-pullela-aws:
### Description
Fixed the Version based Tool Filtering Bug #72 by gracefully handling the version check failure. 
If the Version check fails, the server will not fail anymore, it will expose all tools.

Created a priority order to get the AWS region from different variables and profiles

### Issues Resolved
#72

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-- Handle Tool Filtering failure gracefully and define priority to the AWS Region definitions ([#74](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/74))
 
 ### Added
 - Extend list indices tool ([#68](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/68))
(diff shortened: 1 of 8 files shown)


src/opensearch/helper.py
```diff
@@ -46,12 +50,12 @@ def get_opensearch_version(args: baseToolArgs) -> Version:
     Returns:
         Version: The version of OpenSearch cluster (SemVer style)
     """
-    from .client import initialize_client, is_serverless
-
-    if is_serverless(args):
```
nathaliellenaa, line 67: Should we keep this check for serverless so it doesn't throw an exception when calling `client.info()`?

src/tools/utils.py
```diff
@@ -16,6 +16,9 @@ def is_tool_compatible(current_version: Version, tool_info: dict = {}):
     Returns:
         bool: True if the tool is compatible, False otherwise
     """
+    # Find a version equivalent in serverless mode
+    if not current_version:
```
nathaliellenaa, line 20: This means if it's a serverless cluster or the call to `client.info()` fails in `get_opensearch_version`, we will assume that tool is compatible, right?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+- Handle Tool Filtering failure gracefully and define priority to the AWS Region definitions ([#74](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/74))
```
nathaliellenaa: Nit: move this to a new section called `Changed`/`Improved`?

nathaliellenaa (commented):
Thanks Rithin!! Overall looks good, left a few comments

src/opensearch/helper.py
rithinpullela: We expect it to throw an error which will be caught and would lead to the skipping of tool filtering

src/tools/utils.py
rithinpullela: Yes

CHANGELOG.md
rithinpullela: Will do, thanks for pointing it out!

CHANGELOG.md
```diff
@@ -9,11 +9,15 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 
 ### Fixed
-- Fix endpoint selection bug in ClusterHealthTool and CountTool ([#59](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/59))
-- Fix Serverless issues ([#61](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/61))
+- Handle Tool Filtering failure gracefully and define priority to the AWS Region definitions ([#74](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/74))
 
 ### Security
 
+## [Released 0.2]
```
nathaliellenaa: Seems like there are two `[Released 0.2]` sections (line 16 and 21)?

rithinpullela: Updated to 0.2.2

nathaliellenaa (approved):
Thanks @rithin-pullela-aws, LGTM!

PR #78, opened 2025-07-23
Added new tools for comprehensive OpenSearch Ops analysis

aswath86:
### Description
Adding the following new tools for comprehensive OpenSearch operational analysis,

* GetClusterStateTool
* GetSegmentsTool
* CatNodesTool
* GetNodesTool
* GetIndexInfoTool
* GetIndexStatsTool
* GetQueryInsightsTool
* GetNodesHotThreadsTool
* GetAllocationTool
* GetLongRunningTasksTool

### Issues Resolved
The goal is to create another tools that checks against OpenSearch cluster for best practices and provide tuning recommendations. The Best practices checklist needs a lot of metrics, some/many of which will be obtained from the above API calls.

With just the above tools I'm already able to ask questions such as below using Amazon Q CLI and get reasonably good insights from the OpenSearch cluster.

* can you check if my OpenSearch is following the sharding strategy best practices? 
   * It used ClusterHealthTool, ListIndexTool, GetAllocationTool, CatNodesTool, GetIndexInfoTool, GetShardsTool
* does my OpenSearch have any node hotspots?
   * It used CatNodesTool, GetAllocationTool, GetIndexStatsTool, GetLongRunningTasksTool 
* Are there any hot queries that are consuming a lot of resources?
   * It used GetQueryInsightsTool, GetIndexStatsTool, GetSegmentsTool

Output snippet from the 1st question about sharding best practices

image

   

   
### DIFF
--- a/src/tools/tools.py
+++ b/src/tools/tools.py
@@ -68,7 +68,7 @@ async def list_indices_tool(args: ListIndicesArgs) -> list[dict]:
     try:
         check_tool_compatibility('ListIndexTool', args)
 
-        # If index is provided, get detailed information for that specific index
+        # If index is provided, always return detailed information for that specific index
         if args.index:
             index_info = get_index(args)
             formatted_info = json.dumps(index_info, indent=2)
@@ -76,11 +76,21 @@ async def list_indices_tool(args: ListIndicesArgs) -> list[dict]:
                 {'type': 'text', 'text': f'Index information for {args.index}:\n{formatted_info}'}
             ]
 
-        # Otherwise, list all indices with full information
+        # Otherwise, list all indices
         indices = list_indices(args)
-        formatted_indices = json.dumps(indices, indent=2)
 
-        # Return in MCP expected format
+        # If include_detail is False, return only pure list of index names
+        if not args.include_detail:
+            index_names = [
+                item.get('index')
+                for item in indices
+                if isinstance(item, dict) and 'index' in item
+            ]
+            formatted_names = json.dumps(index_names, indent=2)
+            return [{'type': 'text', 'text': f'Indices:\n{formatted_names}'}]
+
+        # include_detail is True: return full information
+        formatted_indices = json.dumps(indices, indent=2)
         return [{'type': 'text', 'text': f'All indices information:\n{formatted_indices}'}]
     except Exception as e:
         return [{'type': 'text', 'text': f'Error listing indices: {str(e)}'}]
@@ -472,7 +482,7 @@ async def get_long_running_tasks_tool(args: GetLongRunningTasksArgs) -> list[dic
 TOOL_REGISTRY = {
     'ListIndexTool': {
         'display_name': 'ListIndexTool',
-        'description': 'Lists all indices in the OpenSearch cluster with full information including docs.count, docs.deleted, store.size, etc. If an index parameter is provided, returns detailed information about that specific index.',
+        'description': 'Lists indices in the OpenSearch cluster. By default, returns a filtered list of index names only to minimize response size. Set include_detail=true to return full metadata from cat.indices (docs.count, store.size, etc.). If an index parameter is provided, returns detailed information for that specific index including mappings and settings.',
         'input_schema': ListIndicesArgs.model_json_schema(),
         'function': list_indices_tool,
         'args_model': ListIndicesArgs,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 - Add new operational tools for comprehensive OpenSearch cluster analysis: `GetClusterStateTool`, `GetSegmentsTool`, `CatNodesTool`, `GetNodesTool`, `GetIndexInfoTool`, `GetIndexStatsTool`, `GetQueryInsightsTool`, `GetNodesHotThreadsTool`, `GetAllocationTool`, and `GetLongRunningTasksTool` and test cases (#78)
 
+- Add include_detail as optional parameter to ListIndexTool ([#97](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/97))
+
 ### Fixed
 
 ### Removed
--- a/src/tools/tool_params.py
+++ b/src/tools/tool_params.py
@@ -18,6 +18,10 @@ class ListIndicesArgs(baseToolArgs):
         default='',
         description='The name of the index to get detailed information for. If provided, returns detailed information about this specific index instead of listing all indices.',
     )
+    include_detail: bool = Field(
+        default=True,
+        description='Whether to include detailed information. When listing indices (no index specified), if False, returns only a pure list of index names. If True, returns full metadata. When a specific index is provided, detailed information (including mappings) will be returned.',
+    )
 
 
 class GetIndexMappingArgs(baseToolArgs):
--- a/tests/tools/test_tools.py
+++ b/tests/tools/test_tools.py
@@ -111,8 +111,8 @@ class TestTools:
         self.init_client_patcher.stop()
 
     @pytest.mark.asyncio
-    async def test_list_indices_tool(self):
-        """Test list_indices_tool returns full JSON info for all indices."""
+    async def test_list_indices_tool_default_full(self):
+        """Default behavior: returns full JSON info for all indices (include_detail=True)."""
         # Setup: mock full index info as returned by OpenSearch cat.indices
         self.mock_client.cat.indices.return_value = [
             {
@@ -145,7 +145,7 @@ class TestTools:
         # Assert
         assert len(result) == 1
         assert result[0]['type'] == 'text'
-        # Should include the full JSON output
+        # Should include the full JSON output by default
         assert '"index": "index1"' in result[0]['text']
         assert '"docs.count": "100"' in result[0]['text']
         assert '"index": "index2"' in result[0]['text']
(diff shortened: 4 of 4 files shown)


ssunno:
And here's concern about the tools:

In this PR, many tools being proposed. It is known that having too many tools can cause several issues.

* https://www.reddit.com/r/mcp/comments/1jimv3u/whatll_hapen_if_there_has_a_lots_of_tool_in_mcp
* https://blog.sshh.io/p/everything-wrong-with-mcp

Instead of exposing all these tools, it might be better to group theses tools with dynamic toolset management approach described in [here](https://mcpcat.io/blog/mcp-server-best-practices/). What do you think?

FYI) The `listTemplate` proposed in the RFC #83 provides search functionality through meta tooling (or dynamic tool?), rather than exposing various search features as individual tools.

rithin-pullela-aws:
Thanks for creating the PR @aswath86 !!

While these tools do expand the use cases of the MCP Server, as @ssunno mentioned this could degrade the performance of the LLMs as well because of tool explosion.

I propose a way where we can allow cx uploading in #84, while it is still a draft version, I would appreciate it if you can take a look and let us know if it can solve your use case.

rithin-pullela-aws:
I think these are the main alternatives that we can consider:
- Leverage Custom tools from #84 
- Find a way to group these tools into fewer tools as proposed by @ssunno above

aswath86:
> how does it know the best practices? Does the LLM become an OpenSearch expert by being given prior knowledge or by performing web searches? Or does it just rely on its own internal knowledge?

Two part answer. 1/ Most SOTA LLMs are already OpenSearch-knowledgable to use the information from these tools and answer questions and provide recommendations towards best practices. The attached screenshot is a good example of that. All the test I ran with AmazonQ CLI Claude4 had valid-good response.  2/ As already mentioned above, I plan on to add another tool to this that has the best practices checklist for various topics (eg., sharding strategy, mapping optimizations, query tuning etc., etc.,) which will act as a supplement to LLMs knowledge and should help the LLMs to be a bit more systematic.


> In this PR, many tools being proposed. It is known that having too many tools can cause several issues

Agreed. We then have to first define the purpose of this MCP. My understanding of this MCP is that it should have foundational tools in it and not specalized tools. With the currently available tools, I can do few independent task but I don't see full comprehensive use-cases in it. The few more tools I'm adding here gives it a use-case/purpose (plus the individual independent task that they are capable off). While I do agree to not add several tools, which is very valid if, for example, the tool is trying correlate logs, metrics, traces for an observability use-case. As observability is a pillar of its own and not a foundational tool like the above ones but a specialized tool, such tools shouldn't be here. This [es-apm](https://github.com/byviz/mcp-server-elasticsearch-ai) MCP is one such example with specalized tools. Its a separate MCP. Another OpenSearch MCP could be for time-series data with tools actively looking for alerts and Anomaly Detection. 

> FYI) The listTemplate proposed in the RFC https://github.com/opensearch-project/opensearch-mcp-server-py/issues/83 provides search functionality through meta tooling (or dynamic tool?), rather than exposing various search features as individual tools.

May not be a good analogy. This RFC proposes breaking about the `_search` endpoint but the above tools are not breakdowns of a single component such as `_search`. But I will go over the RFC again, later.


> While these tools do expand the use cases of the MCP Server, as @ssunno mentioned this could degrade the performance of the LLMs as well because of tool explosion.

All the test that I did with AmazonQ CLI Claude4, it was intelligent enough to be able to use only the necessary tools for the given task. The context window did overflow when I asked for a generic task such as "Is my OpenSearch cluster following all the best practices" as it tried to use several tools. To address this is why I'm thinking of this other tools that will carry the best practices checklist, providing a systemic approach to the LLM and controlling the tool calls. But until then (or even after that), by following proper prompt etiquette and asking for specific task, I don't see why the LLM performance would degrade. Some examples from my test are below, which shows how Q CLI was able to use only the necessary tools for the given task,

>> * can you check if my OpenSearch is following the sharding strategy best practices? 
>>    * It used ClusterHealthTool, ListIndexTool, GetAllocationTool, CatNodesTool, GetIndexInfoTool, GetShardsTool
>>
>> * does my OpenSearch have any node hotspots?
>>    * It used CatNodesTool, GetAllocationTool, GetIndexStatsTool, GetLongRunningTasksTool 
>>
>> * Are there any hot queries that are consuming a lot of resources?
>>    * It used GetQueryInsightsTool, GetIndexStatsTool, GetSegmentsTool

&nbsp;
&nbsp;
&nbsp;

>> too many tools

Also, "too many tools" of tools is subjective. OpenSearch is really wide software, so 15 may not be "too many" Vs. 15 would be "too many" for a much smaller application. Hence I believe we will have to start thinking about have foundational tools in here and specalized tools in either a separate MCP or using the [Create Custom tools via the Config file](https://github.com/opensearch-project/opensearch-mcp-server-py/issues/84)

ssunno:
> Also, "too many tools" of tools is subjective. OpenSearch is really wide software, so 15 may not be "too many" Vs. 15 would be "too many" for a much smaller application.

Agreed. However, the tools you implemented in this PR are quite diverse, and I think this strategy may expand further in the future. For example, `get_long_running_tasks`, `get_allocation`, `get_segments`, `get_nodes`, and `get_shards` are all subdomains of the cat API. It would be better to integrate these queries into a single `_cat` tool and allow users to select the specific function through Args. What do you think?

ssunno:
> The context window did overflow when I asked for a generic task such as "Is my OpenSearch cluster following all the best practices" as it tried to use several tools.

I’ve experienced a similar issue. I believe this happens because our tools return the OpenSearch API responses without any processing. 
For example, in my short test, the `_cluster/state` API called by the `GetClusterStateTool` returned 62,987 characters even on a single-node OpenSearch with 15 indices. In a production cluster with many more nodes and indices, this API could return over a million chars.

In my opinion, we should refine the tool responses. OpenSearch APIs returns various information, but I don't think we need all of them to diagnose OpenSearch status. In MCP, it might be better to combine and clean up several statistics APIs, then provide only the relevant data to the LLM based on the domain being investigated.
We could create separate tools like `ResourceStatisticsTool` for resource status like node stats, and `DataStatisticsTool` for index & shards. Or merge them into a single tool `StatisticsTool` then LLM can choose data it wants via arguments.

aswath86:
> the tools you implemented in this PR are quite diverse

I still not sure why "diverse" is a problem. Especially when we need diverse set of tools to address foundational use-cases such as cluster management and cluster optimization. 

> It would be better to integrate these queries into a single _cat tool and allow users to select the specific function through Args. What do you think?

Yes, I like that idea. Let me give it some thoughts about this, as each of the `_cat` call has different arguments. If this is implemented then do you propose to remove the already existing tools such as `get_shards`? Also, once again this would make me go back to the topic of "what is the purpose and scope of this OpenSearch MCP" as `_cat` is a quite wide. 

> in my short test, the _cluster/state API called by the GetClusterStateTool returned 62,987 characters even on a single-node OpenSearch with 15 indices. In a production cluster with many more nodes and indices, this API could return over a million chars.
> In my opinion, we should refine the tool responses. OpenSearch APIs returns various information, but I don't think we need all of them to diagnose OpenSearch status

Also agreed. For this, I'm hoping to restrict the tools response by only asking for (restricted _CAT API call) what is required for the task, by creating this `BestPracticesTool`?. Basically the `BestPracticesTool` is going to have prompts that will guide the LLM to only run APIs with arguments restricting and getting a learner response.  

This `BestPracticesTool` will also act as supplementary knowledge for LLMs so the task can be much more systematic. 

@ssunno - We could discuss this over a call on Slack. If you are able to find me on the OpenSearch Slack could you ping me so we find some time to go over this, please. Thank you!

Jon-AtAWS:
@aswath86 and I caught up over a phone call. 

To my mind, there are two spheres of interest: application, and operations. When using OpenSearch as part of a larger, AI-powered system, you want to query and retrieve information in OpenSearch indices (e.g., for RAG). When you are operating OpenSearch, you want to use operational data as part of an AI assistant. There is some gray area - you might want to know all of the indices and their contents as part of an application's data discovery. You might also want to know the same for understanding the resource utilization. But mostly I think application and operations are distinct use cases, and shouldn't live in the same MCP server.

A bigger question is whether we should decide to maintain multiple MCP servers in the project, or whether it's better to have these servers in multiple, self-maintained repos, outside of the project. There will certainly be other repos with MCP servers for OpenSearch, and multiple, open-source MCP servers for many different functions and data stores. 

Net-net, I support adding a different MCP server/repo for the operational case to the project. With that, I think the project covers the necessary bases. Other use cases can be hosted in external repos.

src/tools/tools.py
```diff
@@ -153,4 +503,94 @@ async def get_shards_tool(args: GetShardsArgs) -> list[dict]:
...
+        'min_version': '1.0.0',
+        'http_methods': 'GET',
+    },
+    'GetQueryInsightsTool': {
+        'display_name': 'GetQueryInsightsTool',
+        'description': 'Gets query insights from the /_insights/top_queries endpoint, showing information about query patterns and performance.',
+        'input_schema': GetQueryInsightsArgs.model_json_schema(),
+        'function': get_query_insights_tool,
+        'args_model': GetQueryInsightsArgs,
+        'min_version': '2.0.0',  # Query insights feature requires OpenSearch 2.0+
```
jiapingzeng: Looks like query insights requires 2.12+
https://docs.opensearch.org/latest/observing-your-data/query-insights/index/

src/tools/tools.py
```diff
@@ -153,4 +503,94 @@ async def get_shards_tool(args: GetShardsArgs) -> list[dict]:
...
+        'description': 'Gets information about long-running tasks in the cluster, sorted by running time in descending order.',
+        'input_schema': GetLongRunningTasksArgs.model_json_schema(),
+        'function': get_long_running_tasks_tool,
+        'args_model': GetLongRunningTasksArgs,
+        'min_version': '1.0.0',
+        'http_methods': 'GET',
+    },
+    'GetNodesTool': {
+        'display_name': 'GetNodesTool',
+        'description': 'Gets detailed information about nodes in the OpenSearch cluster, including static information like host system details, JVM info, processor type, node settings, thread pools, installed plugins, and more. Can be filtered by specific nodes and metrics.',
```
jiapingzeng, line 599: would suggest differentiating the descriptions of `CatNodesTool` and `GetNodesTool` more, maybe provide example use cases? I'm not sure if LLM will be able to understand the difference between them.

src/tools/tools.py
aswath86: Ah right.. Fixed.

src/tools/tools.py
aswath86: Gave a better description for `CatNodesTool`. Tested with AmazonQ CLI Claude4.0 and it is able to use the correct tool for the given task. Below were the test. 

- what is the available processors on each of my OpenSearch nodes?
   - It used `GetNodesTool`
- are there any resource usage skew in my OpenSearch cluster??
  - It used `CatNodesTool`, `GetAllocationTool`
   
The answers and recommendations provided using these tools were quite reasonable.

rithin-pullela-aws:
@aswath86 can you resolve conflicts?

PR #80, opened 2025-07-24
Fix Tool Renaming Edge cases

rithinpullela:
### Description
- Enforces Tool names to follow `^[a-zA-Z0-9_-]+$` Regex pattern
- Prioritize Config file over CLI arguments (to follow the same pattern as tool filtering)
- Throw error when 2 tools are named to the same value
- Throw error when a tool is given conflicting values in aliases
-- Example: 
 ```
 ListIndexTool:
   displayName:abc
   name:xyz
```
- Add comprehensive testing for the cases
### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Handle Tool Filtering failure gracefully and define priority to the AWS Region definitions ([#74](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/74))
+- Fix Tool Renaming Edge cases ([#80](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/80))
 
 ### Security
 

### CODE REVIEW

File: src/tools/config.py [hunk 1]
```diff
@@ -14,84 +14,249 @@
...
+                    raise ValueError(
+                        f"Duplicate description field for tool '{original_tool_name}' in config file. "
+                        f'Found multiple aliases: {[k for k, v in custom_config.items() if _find_actual_field(k) == DESCRIPTION_STRING]}'
+                    )
+                file_configs[original_tool_name][DESCRIPTION_STRING] = config_value
+                set_fields.add(DESCRIPTION_STRING)
+            else:
+                logging.warning(
+                    f"Invalid field '{config_key}' for tool '{original_tool_name}' in config file will be ignored. "
+                    f'Only display_name and description are supported.'
```
nathaliellenaa, line 103: This means other name like `customName` or `customDescription` are not supported? But it's mentioned in the user guide that other names are supported https://github.com/opensearch-project/opensearch-mcp-server-py/blob/7abbf41801e6b82b46775bbd5fca096e62b38e85/USER_GUIDE.md#supported-field-aliases

src/tools/config.py
```diff
@@ -118,16 +287,23 @@ def apply_custom_tool_config(
...
-    # Apply configurations (CLI overrides file configs)
-    all_configs = file_configs + cli_configs
-    _apply_validated_configs(custom_registry, all_configs)
+    # Load configurations from appropriate source
+    if config_from_file:
+        # Use config file and completely ignore CLI
+        file_configs = _load_config_from_file(config_from_file)
+        if file_configs:
+            _validate_config(file_configs)
+            _apply_validated_configs(custom_registry, file_configs)
```
nathaliellenaa, line 298: For config file, it first checks for duplicate name within the config file itself inside `_load_config_from_file` function  https://github.com/opensearch-project/opensearch-mcp-server-py/blob/7abbf41801e6b82b46775bbd5fca096e62b38e85/src/tools/config.py#L85-L89 then it checks for duplicate name with the existing tool registry inside `_validate_config`
https://github.com/opensearch-project/opensearch-mcp-server-py/blob/7abbf41801e6b82b46775bbd5fca096e62b38e85/src/tools/config.py#L229-L232 Am I understanding this correctly?

src/tools/config.py
```diff
@@ -14,84 +14,249 @@
...
-        if actual_field is None:
-            # Log warning for unsupported field
-            logging.warning(f"Warning: Unsupported field '{field_alias}' for tool '{tool_id}'.")
-            continue
-        # Special handling for description length validation
-        if actual_field == 'description' and len(field_value) > 1024:
-            logging.warning(
-                f"Warning: The description for '{field_alias}' exceeds 1024 characters ({len(field_value)}). "
-                f"Some LLM models may not support long descriptions."
-            )
```
nathaliellenaa, line 53: Should we keep this description length validation?

nathaliellenaa (commented):
Overall looks good, left some small comments

src/tools/config.py
rithinpullela: These aliases will be supported. But an invalid alias will be ignored. Like 
```
tools:
  ListIndexTool:
    display_name: "Index_Manager"
    description: "List and manage OpenSearch indices"
    random: "rand value"
```

The `random` field will be ignored

src/tools/config.py
rithinpullela: No, we don't need to because we don't see the bug with OpenAI models anymore

src/tools/config.py
rithinpullela: These are 2 different scenarios:
 First one is when the cx provides 2 aliases for the tool name. 
 example:
```
tools:
  ListIndexTool:
    display_name: "Index_Manager"
    description: "List and manage OpenSearch indices"
    name: "rand_name"
```
Now we have conflict for the actual name and hence we raise an error.

The second one is reusing the names.
Example:
```
tools:
  ListIndexTool:
    display_name: "MY_SAME_NAME"
    description: "List and manage OpenSearch indices"
  IndexMappingTool:
    display_name: "MY_SAME_NAME"
    description: "List and manage OpenSearch indices"
```
Since both tools are trying to set the same name, we are throwing an error

src/tools/config.py
nathaliellenaa: Got it, thanks for the clarification!

src/tools/config.py
nathaliellenaa: Ah ok, that makes a lot of sense

PR #100, opened 2025-08-21
add tool args description modify feature to config file parser

### LINKED ISSUE #75: [FEATURE] Support modify the description of Tool arguments
In #9 , we added the feature to update the tool name and description. It is good to support customizing the tool Args too.

If users can modify the Args, they can provide hints for their purpose when the AI agent generates queries.

User can add all additional information to the description field, which already supports customization. However, I believe this would be a less intuitive hint for the AI agent.

ssunno:
### Description

This PR introduces support for editing not only the tool name and description, but also the argument descriptions via the `config.yaml` file.
A new args field can now be added in config.yaml. If invalid args are provided, a ValueError will be raised.

In config.yaml

~~~ yaml
tools:
  ListIndexTool:
    display_name: "CustomIndexTool"
    description: "This is a custom description"
    args:
      index: "Custom description for the 'index' argument in ListIndexTool."
~~~

UPDATE: Users can also modify the args description using CLI runtime parameters. For example,

~~~
python -m mcp_server_opensearch --transport stream  --tool.GetShardsTool.args.index.description="Custom descriptions by ssunno" 
~~~

### Issues Resolved
Closes #75


ssunno:
@nathaliellenaa Updated. :)

nathaliellenaa (approved):
Thanks @ssunno, LGTM!

PR #108, opened 2025-10-02
Add alias and data stream to list tool

### LINKED ISSUE #56: [FEATURE] Return alias and datastream in ListIndexTool

I think `ListIndexTool` should also return `aliases` and `data streams`. In OpenSearch, data can be queried not only from indices, but also from aliases and data streams.

Here’s why I prefer including them in ListIndexTool rather than creating separate tools:
1. Having multiple tools that look similar may confuse the LLM.
2. Aliases and data streams have backing indices, which could already appear in the list returned by ListIndexTool. However, querying the backing indices directly may go against the original design intent.
3. Managing them together allows us to clearly distinguish between indices, aliases, and data streams. It also helps reduce duplication and better reflect the intended index design.

`ListIndexTool` returns `aliases` and `data stream`. Eliminate backing indices from the `_cat/indices`.

Make `ListAliasTool`, `ListDatastreamTool`. But it look similar with `ListIndexTool` and confused.

> I'm not sure this is really useful. Please feel free comment if you have better idea !

ssunno:
### Description

This PR add alias and data stream to return of `ListIndicesTool`. 

The `aliases` and `data streams` are concepts that group multiple indices. But current version of `ListIndicesTool` does not support them, so it returns all indices equally, exposing the backing indices of data streams directly. To apply the intention of index designers, aliases and data streams should be returned as separate groups.

Note: Indices that belong to an alias or a data stream are excluded from the general index list to avoid duplication.

When `include_detail = False`, aliases and data streams return only the number of indices instead of their names.

example of `ListIndicesTool` (include_detail = False):

~~~
Indices, aliases, and data streams information:
{
  "indices": [
    "sonested",
    ".opendistro_security",
    ".opensearch-observability",
    ".plugins-ml-config"
  ],
  "aliases": [
    {
      "alias": "logs_current",
      "index_count": 2
    }
  ],
  "data_streams": [
    {
      "name": "logs-ds",
      "timestamp_field": {
        "name": "@timestamp"
      },
      "generation": 1,
      "status": "YELLOW",
      "template": "logs-template",
      "index_count": 1
    }
  ]
}
~~~

### Issues Resolved
Resolves #56 

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -3,7 +3,9 @@
 
 import json
 import logging
+from copy import deepcopy
 from semver import Version
+from opensearchpy.exceptions import NotFoundError
 from tools.tool_params import *
 
 # Configure logging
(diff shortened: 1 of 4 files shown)


src/opensearch/helper.py
```diff
@@ -12,12 +14,104 @@
...
-def list_indices(args: ListIndicesArgs) -> json:
+def list_indices(args: ListIndicesArgs) -> list[dict]:
     from .client import initialize_client
 
     client = initialize_client(args)
     response = client.cat.indices(format='json')
-    return response
+    if not isinstance(response, list):
+        logger.warning('Unexpected response type for cat.indices: %s', type(response))
+        return []
```
rithinpullela, line 24: If the user is not authorized to perform the request, we might get an error response. This error should be propagated to the client making request in some form. But this would return an empty list

src/opensearch/helper.py
```diff
@@ -12,12 +14,104 @@
...
+    client = initialize_client(args)
+
+    try:
+        response = client.indices.get_alias(index='*')
+    except NotFoundError:
+        return {}
+
+    if not isinstance(response, dict):
+        logger.warning('Unexpected response type for indices.get_alias: %s', type(response))
+        return {}
```
rithinpullela, line 50: Same as above, should we handle it differently?

src/opensearch/helper.py
```diff
@@ -12,12 +14,104 @@
...
+    client = initialize_client(args)
+
+    try:
+        response = client.indices.get_data_stream(name='*')
+    except NotFoundError:
+        return []
+
+    if not isinstance(response, dict):
+        logger.warning('Unexpected response type for indices.get_data_stream: %s', type(response))
+        return []
```
rithinpullela, line 79: Same as above

src/opensearch/helper.py
```diff
@@ -12,12 +14,104 @@
 
 # List all the helper functions, these functions perform a single rest call to opensearch
 # these functions will be used in tools folder to eventually write more complex tools
-def list_indices(args: ListIndicesArgs) -> json:
+def list_indices(args: ListIndicesArgs) -> list[dict]:
```
rithinpullela, line 17: Nit:
Would not block this PR, but something to perhaps think of

On a high level for list_indices, list_aliases, and list_data_streams

Would it make sense to create a class for the responses and parse the response into the classes? We can perhaps throw some sort of error if parsing fails. That way we can skip doing checks like isInstance etc?

rithinpullela (commented):
Thanks for raising the PR @ssunno, I believe this is a step in the right direction. Added few comments.

rithin-pullela-aws:
Hi @ssunno are you still planning to work on this PR?

rithinpullela:
Hi @ssunno do you plan to work on this PR? If not I can add commits to this PR and merge it. Thank you

ssunno:
@rithinpullela Sorry for delaying. Please don't wait me; feel free to work comfortably. I will participate again elsewhere later.

PR #112, opened 2025-10-09
feat: Add Generic OpenSearch API Tool

### LINKED ISSUE #109: [FEATURE]Generic tool which supports all OpenSearch API
In the current model, a MCP tool is created for each OpenSearch API. This model results in a tool explosion and increases context size. This approach also limits the number of tools that could be supported limiting the capability of the MCP server.

Creating a single general purpose tool which can work with all OpenSearch API.

With the newer LLM models, it is already trained with opensearch documentation and understands various opensearch API and request/response models. We can leverage model knowledge to build a generic tool. The tool will take in URL path, query parameters and request body as input. Tool description should instruct LLM to make use of its opensearch knowledge to invoke the tool.

* Need to check if this approach increases hallucination because it is more open ended.
* Generic tool can be a fallback to existing tools.

### Do you have any additional context?
A similar approach is used by AWS CLI MCP tool: https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py

arjunkumargiri:
Implements a flexible, general-purpose tool that can interact with any OpenSearch API endpoint, addressing the tool explosion problem mentioned in issue #109.

Key Features:
- Single tool interface for all OpenSearch APIs
- Support for all HTTP methods (GET, POST, PUT, DELETE, HEAD, PATCH)
- Flexible parameters: custom paths, query parameters, request bodies, headers
- Write operation protection via OPENSEARCH_SETTINGS_ALLOW_WRITE environment variable
- Leverages LLM knowledge of OpenSearch APIs for intelligent usage
- Same authentication and connection logic as existing tools

Security:
- Write operations (POST, PUT, DELETE, PATCH) blocked when OPENSEARCH_SETTINGS_ALLOW_WRITE=false
- Clear error messages when operations are not allowed
- Runtime permission checking for each request

Benefits:
- Reduces tool explosion and context size
- Increases server capability to handle all OpenSearch APIs
- Future-proof for new API endpoints
- Maintains backward compatibility with existing tools

Files Added:
- src/tools/generic_api_tool.py: Core implementation
- tests/tools/test_generic_tool.py: Functionality tests
- tests/tools/test_write_protection_simple.py: Security tests

Files Modified:
- README.md: Added documentation for new tool
- src/tools/tools.py: Added tool to registry
- src/tools/tool_filter.py: Added to core_tools category

Closes #109

### Testing
Added to tool to a chat agent and tested new tool

Screenshot 2025-10-09 at 4 03 25 PM

### Issues Resolved
Closes #109

### DIFF
--- a/src/tools/generic_api_tool.py
+++ b/src/tools/generic_api_tool.py
@@ -3,6 +3,7 @@
 
 import json
 import logging
+import os
 from typing import Any, Dict, Optional
 from urllib.parse import urlencode
 
@@ -10,7 +11,6 @@ from .tool_params import baseToolArgs
 from pydantic import BaseModel, Field
 
 logger = logging.getLogger(__name__)
-
 
 class GenericOpenSearchApiArgs(baseToolArgs):
     """Arguments for the generic OpenSearch API tool."""
@@ -28,7 +28,7 @@ class GenericOpenSearchApiArgs(baseToolArgs):
     )
     body: Optional[Any] = Field(
         default=None,
-        description='Request body for POST/PUT requests. Can be a JSON object, string, or None'
+        description='Request body for GET/POST/PUT requests. Can be a JSON object, string, or None'
     )
     headers: Optional[Dict[str, str]] = Field(
         default=None,
@@ -85,8 +85,8 @@ async def generic_opensearch_api_tool(args: GenericOpenSearchApiArgs) -> list[di
     The tool supports all HTTP methods and can handle query parameters, request bodies,
     and custom headers. It uses the same authentication and connection logic as other tools.
     
-    Write operations (POST, PUT, DELETE, PATCH) are only allowed when the 
-    OPENSEARCH_SETTINGS_ALLOW_WRITE environment variable is set to "true".
+    Write operations (POST, PUT, DELETE, PATCH) are only allowed when write 
+    operations are enabled via configuration.
     
     Args:
         args: GenericOpenSearchApiArgs containing the API request details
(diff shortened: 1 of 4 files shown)


src/tools/tools.py
```diff
@@ -603,4 +606,13 @@ async def get_long_running_tasks_tool(args: GetLongRunningTasksArgs) -> list[dic
         'min_version': '1.0.0',
         'http_methods': 'GET',
     },
+    'GenericOpenSearchApiTool': {
+        'display_name': 'GenericOpenSearchApiTool',
+        'description': 'A flexible tool for calling any OpenSearch API endpoint. Supports all HTTP methods with custom paths, query parameters, request bodies, and headers. Write operations (POST, PUT, DELETE, PATCH) require OPENSEARCH_SETTINGS_ALLOW_WRITE=true. Use this when you need to access OpenSearch APIs that don\'t have dedicated tools, or when you need more control over the request. Leverages your knowledge of OpenSearch API documentation to construct appropriate requests.',
```
rithinpullela: `Write operations (POST, PUT, DELETE, PATCH) require OPENSEARCH_SETTINGS_ALLOW_WRITE=true`

This description will be provided to the LLM and the LLM cannot use this information to perform an action. We can add this in error message when an un allowed method is called.

src/tools/generic_api_tool.py
```diff
@@ -0,0 +1,167 @@
...
+        default='GET',
+        description='HTTP method to use (GET, POST, PUT, DELETE, HEAD, PATCH)'
+    )
+    query_params: Optional[Dict[str, Any]] = Field(
+        default=None,
+        description='Query parameters to include in the request URL as key-value pairs'
+    )
+    body: Optional[Any] = Field(
+        default=None,
+        description='Request body for POST/PUT requests. Can be a JSON object, string, or None'
```
rithinpullela: nit: could be used for GET too, ex: OpenSearch search API

src/tools/generic_api_tool.py
```diff
@@ -0,0 +1,167 @@
...
+        from opensearch.client import initialize_client
+        
+        # Validate method
+        valid_methods = ['GET', 'POST', 'PUT', 'DELETE', 'HEAD', 'PATCH']
+        method = args.method.upper()
+        if method not in valid_methods:
+            return [{'type': 'text', 'text': f'Error: Invalid HTTP method "{args.method}". Valid methods are: {", ".join(valid_methods)}'}]
+        
+        # Check if write operations are allowed
+        allow_write = os.getenv('OPENSEARCH_SETTINGS_ALLOW_WRITE', 'true').lower() == 'true'
```
rithinpullela: allow write can be set from the config file as well, we need to check that here too.

rithin-pullela-aws (commented):
This would be a great addition, thanks for the PR Arjun. Added 2 small comments

rithin-pullela-aws (approved):
LGTM!
Thanks for the changes!

PR #114, opened 2025-10-15
feat: Add pagination support to SearchIndexTool and GetSegmentsTool to prevent token overflow

brunoselvacj:
### Description
  This change adds pagination support to SearchIndexTool and GetSegmentsTool to prevent token overflow errors when MCP tool responses exceed the 25,000 token limit.

  **SearchIndexTool changes:**
  - Added `size` parameter (default: 10, max: 100) to limit number of search results returned
  - Added `from` parameter (default: 0) to support offset-based pagination
  - Automatic capping at 100 results to prevent token overflow
  - 7 new comprehensive tests covering pagination scenarios

  **GetSegmentsTool changes:**
  - Added `limit` parameter (default: 1000) to cap number of segments returned
  - Prevents token overflow for clusters with many segments
  - 3 new tests covering limit functionality

  **Additional improvements:**
  - Updated README.md with new pagination parameters
  - Updated tool descriptions in TOOL_REGISTRY
  - All 50 tests passing with TDD approach (wrote failing tests first, then implemented features)
  - Maintains backward compatibility with default values

  ### Issues Resolved
  Fixes token overflow issues where SearchIndexTool and GetSegmentsTool responses exceed the 25,000 token MCP limit, causing errors like: `MCP tool 'SearchIndexTool' response (35969 tokens) exceeds maximum
  allowed tokens (25000)`.

  For more information on following Developer Certificate of Origin and signing off your commits, please check
  [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -45,10 +45,28 @@ def get_index_mapping(args: GetIndexMappingArgs) -> json:
 
 
 def search_index(args: SearchIndexArgs) -> json:
+    """Search an index with pagination support.
+
+    Args:
+        args: SearchIndexArgs containing index, query, and optional pagination params
+
+    Returns:
+        json: Search results from OpenSearch
+    """
     from .client import initialize_client
 
     client = initialize_client(args)
-    response = client.search(index=args.index, body=args.query)
+
+    # Ensure query is a dict for merging
+    query_body = args.query if isinstance(args.query, dict) else {}
+
+    # Apply pagination parameters (override any user-provided values)
+    # Cap size at maximum of 100 to prevent token overflow
+    effective_size = min(args.size, 100) if args.size else 10
+    query_body['size'] = effective_size
+    query_body['from'] = args.from_ if args.from_ is not None else 0
+
+    response = client.search(index=args.index, body=query_body)
     return response
 
 
@@ -62,21 +80,26 @@ def get_shards(args: GetShardsArgs) -> json:
 
 def get_segments(args: GetSegmentsArgs) -> json:
     """Get information about Lucene segments in indices.
-    
+
     Args:
-        args: GetSegmentsArgs containing optional index filter
-        
+        args: GetSegmentsArgs containing optional index filter and limit
+
     Returns:
         json: Segment information for the specified indices or all indices
     """
     from .client import initialize_client
-    
+
     client = initialize_client(args)
-    
+
     # If index is provided, filter by that index
     index_param = args.index if args.index else None
-    
+
     response = client.cat.segments(index=index_param, format='json')
+
+    # Apply limit to prevent token overflow
+    if args.limit and isinstance(response, list):
+        return response[:args.limit]
+
     return response
 
 
--- a/src/tools/tools.py
+++ b/src/tools/tools.py
@@ -499,7 +499,7 @@ TOOL_REGISTRY = {
     },
     'SearchIndexTool': {
         'display_name': 'SearchIndexTool',
-        'description': 'Searches an index using a query written in query domain-specific language (DSL) in OpenSearch',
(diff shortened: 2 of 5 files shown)


nathaliellenaa:
Thanks for raising this PR @brunoselvacj! Can you add the changes to the changelog? I also have a high level question regarding the `SearchIndexTool` size parameter. I'm wondering whether we should maintain a hard maximum limit of 100 or consider a more flexible approach. An alternative would be to remove the hard limit and instead throw a warning when the size exceeds 100. What do you think about this?

src/opensearch/helper.py
```diff
@@ -45,10 +45,28 @@ def get_index_mapping(args: GetIndexMappingArgs) -> json:
...
-    response = client.search(index=args.index, body=args.query)
+
+    # Ensure query is a dict for merging
+    query_body = args.query if isinstance(args.query, dict) else {}
+
+    # Apply pagination parameters (override any user-provided values)
+    # Cap size at maximum of 100 to prevent token overflow
+    effective_size = min(args.size, 100) if args.size else 10
+    query_body['size'] = effective_size
+    query_body['from'] = args.from_ if args.from_ is not None else 0
```
rithinpullela, line 67: Can we make the max limit variable configurable? Either via CLI parameter or config file? We can default to a maximum of 100 if the variable is not provided

src/opensearch/helper.py
```diff
@@ -62,21 +80,26 @@ def get_shards(args: GetShardsArgs) -> json:
...
     # If index is provided, filter by that index
     index_param = args.index if args.index else None
-    
+
     response = client.cat.segments(index=index_param, format='json')
+
+    # Apply limit to prevent token overflow
+    if args.limit and isinstance(response, list):
+        return response[:args.limit]
+
```
rithinpullela, line 102: When we apply the max limit, perhaps adding the information that the size has been truncated in the response message sent back to the MCP client would help?

The same thing can be applied for the Search Index tool response as well

src/tools/tools.py
```diff
@@ -499,7 +499,7 @@ async def get_long_running_tasks_tool(args: GetLongRunningTasksArgs) -> list[dic
     },
     'SearchIndexTool': {
         'display_name': 'SearchIndexTool',
-        'description': 'Searches an index using a query written in query domain-specific language (DSL) in OpenSearch',
+        'description': 'Searches an index using a query written in query domain-specific language (DSL) in OpenSearch. Supports pagination with size (default: 10, max: 100) and from parameters to limit response size and prevent token overflow.',
```
rithinpullela, line 502: This new information about the parameters is sent to the LLM as parameter description. Maybe we can limit the tool description to just it's function and not describe its parameters?

src/tools/tools.py
```diff
@@ -524,7 +524,7 @@ async def get_long_running_tasks_tool(args: GetLongRunningTasksArgs) -> list[dic
     },
     'GetSegmentsTool': {
         'display_name': 'GetSegmentsTool',
-        'description': 'Gets information about Lucene segments in indices, including memory usage, document counts, and segment sizes. Can be filtered by specific indices.',
+        'description': 'Gets information about Lucene segments in indices, including memory usage, document counts, and segment sizes. Can be filtered by specific indices. Supports limit parameter (default: 1000) to prevent token overflow.',
```
rithinpullela, line 527: same concern as above

rithinpullela (commented):
Overall looks like a positive step to improve reliability of the tools, but added few concerns with the change

rithinpullela:
Looks like this PR is no longer being worked on, closing this for now, please open a new PR if you want to continue your work @brunoselvacj, thanks!

PR #117, opened 2025-10-23
Add header-based authentication + Code Clean up

rithinpullela:
### Header Auth:
See 2nd commit to only look at these changes

This PR adds support for header-based authentication, allowing AWS credentials and OpenSearch connection details to be passed dynamically through HTTP request headers instead of using static environment variables or configuration files.

- **New Environment Variable**: `OPENSEARCH_HEADER_AUTH=true` enables header-based authentication.
- **New Config Variable**: `opensearch_header_auth=true` enables header-based authentication for multi mode.
- Only works with streaming servers (`--transport stream`)
- Not suitable for STDIO mode
- Falls back to env variables or config file variables
- Supports both multi and single mode

### Code clean up to separate single and multi mode:
See 1st commit to only look at these changes

- Multi mode showed cluster name as not mandatory, it changes now.
- Instead of relying on pydantic to validate the required arguments, we throw a custom error
- Single mode and multi mode client creation separated for clarity and readability.
- Global variable to get current mode

### Issues Resolved

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -15,7 +15,7 @@
 
 ## Overview
 
-The OpenSearch MCP (Model Context Protocol) Server  provides a bridge between AI agents and OpenSearch clusters. It supports both single-cluster and multi-cluster configurations with various authentication methods including IAM roles, basic authentication, AWS credentials, and header-based authentication.
+The OpenSearch MCP (Model Context Protocol) Server provides a bridge between AI agents and OpenSearch clusters. It supports both single-cluster and multi-cluster configurations with various authentication methods including IAM roles, basic authentication, AWS credentials, and header-based authentication.
 The server can be started in both STDIO and streaming modes.
 
 ### Streaming Server


jiapingzeng:
High-level question: seems that we accept `opensearch-url` as a header so only single cluster mode is supported. Should we support multi-mode in MCP server layer here or Agent layer?

I think either approach is fine but If we do want to support multi-mode via headers, will also need to take in data source type. For example, if an user wants to use a mix of AOS domains and AOSS collections.

jiapingzeng (approved):
LGTM!

rithin-pullela-aws:
> High-level question: seems that we accept `opensearch-url` as a header so only single cluster mode is supported. Should we support multi-mode in MCP server layer here or Agent layer?
> 
> I think either approach is fine but If we do want to support multi-mode via headers, will also need to take in data source type. For example, if an user wants to use a mix of AOS domains and AOSS collections.


This is a good point JP. So currently we support both single and multi mode. 
In single mode, if headers are used, different connections using different URL variables in the headers can be used to connect to different clusters.

In case of multi mode, the users can choose which clusters should support header based auth via the config file.  

Currently users can choose where to implement the "multi mode" support, no matter what they choose(Agent layer via headers or MCP layer), we support the modes

USER_GUIDE.md
```diff
@@ -15,7 +15,7 @@
 
 ## Overview
 
-The OpenSearch MCP (Model Context Protocol) Server Python provides a bridge between AI agents and OpenSearch clusters. It supports both single-cluster and multi-cluster configurations with various authentication methods including IAM roles, basic authentication, and AWS credentials.
+The OpenSearch MCP (Model Context Protocol) Server  provides a bridge between AI agents and OpenSearch clusters. It supports both single-cluster and multi-cluster configurations with various authentication methods including IAM roles, basic authentication, AWS credentials, and header-based authentication.
```
nathaliellenaa: Nit: remove double spaces between `Server  provides`

src/opensearch/client.py
```diff
@@ -120,29 +128,55 @@ def _initialize_client_single_mode() -> OpenSearch:
...
+
+        # Check if header auth is enabled and update variables accordingly
+        use_header_auth = os.getenv('OPENSEARCH_HEADER_AUTH', '').lower() == 'true'
+        if use_header_auth:
+            header_auth = _get_auth_from_headers()
+            header_url = header_auth.get('opensearch_url')
+            if header_url:
+                opensearch_url = header_url
+            header_service = header_auth.get('aws_service_name')
+            if header_service:
```
nathaliellenaa, line 155: should we throw warning/error for invalid service?

rithinpullela: I think it is okay, we will default to `es` if an invalid request is sent

PR #121, opened 2025-10-30
[FEATURE] Integrate skill tool

PauiC:
### Description
Integrate skill tool

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,13 @@ The following tools are available but disabled by default. To enable them, see t
 - [GetAllocationTool](https://docs.opensearch.org/docs/latest/api-reference/cat/cat-allocation/): Gets information about shard allocation across nodes in the cluster from the /\_cat/allocation endpoint.
 - [GetLongRunningTasksTool](https://docs.opensearch.org/docs/latest/api-reference/cat/cat-tasks/): Gets information about long-running tasks in the cluster, sorted by running time in descending order.
 
+### Skills Tools (Enabled by Default)
+
+Advanced analysis tools for data analysis and troubleshooting.
+
+- [DataDistributionTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/data-distribution-tool/): Analyzes data distribution patterns and field value frequencies within OpenSearch indices. Supports both single dataset analysis and comparative analysis between two time periods to identify distribution changes.
+- [LogPatternAnalysisTool](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/log-pattern-analysis-tool/): Detects anomalous log patterns and sequences through comparative analysis between baseline and selection time ranges. Supports log sequence analysis with trace correlation, log pattern difference analysis, and log insights analysis for error detection.
+
 ### Tool Parameters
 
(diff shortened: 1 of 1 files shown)


src/tools/skills_tools.py
```diff
@@ -0,0 +1,105 @@
...
+        client = initialize_client_with_cluster(None)
+
+        # Call OpenSearch ML tools execute API
+        response = client.transport.perform_request(
+            'POST',
+            f'/_plugins/_ml/tools/_execute/{tool_name}',
+            body={'parameters': parameters}
+        )
+
+        logger.info(f"工具 {tool_name} 结果: {json.dumps(response, indent=2)}")
```
Hailong-am: we can improve the log message by remove Chinese characters

Hailong-am:
can we put some test result for the new tool too?

src/tools/skills_tools.py
PauiC: Thanks for your suggestion. I've resolved it.

src/tools/skills_tools.py
```diff
@@ -0,0 +1,105 @@
...
+    traceFieldName: str = Field(default="", description="Field for trace/correlation ID (optional)")
+    baseTimeRangeStart: str = Field(default="", description="Start time for baseline comparison period (optional)")
+    baseTimeRangeEnd: str = Field(default="", description="End time for baseline comparison period (optional)")
+
+async def call_opensearch_tool(tool_name: str, parameters: Dict[str, Any]) -> List[Dict]:
+    """Call OpenSearch ML tools API"""
+    try:
+        from opensearch.client import initialize_client_with_cluster
+
+        client = initialize_client_with_cluster(None)
```
rithinpullela: This function has been changed in the latest PR. Can you use `initialize_client` from `opensearch.client`?

Also instead of doing lazy import can we have the import statement at the top?

rithinpullela: Also for client initialization you must send the BaseToolArgs params in the `initialize_client()` function call. That will be used to support multi cluster mode

rithin-pullela-aws (commented):
Thanks for the PR. Over all it looks good, added 2 small comments.

Can we add some Unit tests for this?

src/tools/skills_tools.py
PauiC: Thanks for your suggestion. I've resolved it.

Hailong-am (approved):

src/tools/skills_tools.py
```diff
@@ -0,0 +1,104 @@
...
+    """Call OpenSearch ML tools API"""
+    try:
+        client = initialize_client(args)
+
+        # Call OpenSearch ML tools execute API
+        response = client.transport.perform_request(
+            'POST',
+            f'/_plugins/_ml/tools/_execute/{tool_name}',
+            body={'parameters': parameters}
+        )
```
rithinpullela: With the new release we have an async OpenSearch client now which makes the MCP Server concurrent. 

Can we do await here before making the request to the client? Same for all other skills tools as well.

PauiC: Thanks for your suggestion. I've resolved it.

PR #122, opened 2025-11-03
Default size to 100

yyfamazon:
### Description
default size to 100

### Issues Resolved

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ pyrightconfig.json
 *-codecov.xml
 
 # Environments
-.venv
+.venv*
 
 # Python-generated files
 __pycache__/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,15 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 
 ### Fixed
+- Fix AWS auth issues for cat based tools, pin OpenSearchPy to 2.18.0 ([#135](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/135))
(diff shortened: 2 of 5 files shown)


src/opensearch/helper.py
```diff
@@ -48,7 +48,13 @@ def search_index(args: SearchIndexArgs) -> json:
     from .client import initialize_client
 
     client = initialize_client(args)
-    response = client.search(index=args.index, body=args.query)
+    
+    # Set default size to 100 if not present in query body
```
Hailong-am: this seems that not correct to me. If size is not present, default is 10 and then we override it 100?

yyfamazon: resolved, thanks

src/opensearch/helper.py
```diff
@@ -48,7 +48,13 @@ def search_index(args: SearchIndexArgs) -> json:
...
+    query = args.query.copy() if isinstance(args.query, dict) else args.query
+    if isinstance(query, dict) and 'size' in query:
+        query['size'] = min(100, query['size'])
```
rithinpullela: This would mean we would not be able to send a size > 100.

I believe this kind of restriction is not ideal, can we make it configurable in settings?

rithin-pullela-aws (commented):
Add a high level question, can you also add change log and sign your commit to fix DCO?

src/opensearch/helper.py
rithinpullela: Looks like a similar PR has already been raised in #114

yyfamazon: should I close my pr if a similar PR has already been raised ?

PR #126, opened 2025-11-13
[Fix] Close OpenSearch client gracefully

rithinpullela:
### Description
- Close OpenSearch client gracefully using AsyncIterator
- In multi mode, test connection before adding into the allowed clusters is removed because with header based authentication we cannot connect to clusters without headers. And the error is better propagated in run time to the users

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Fixed
 - Fix Concurrency: Use Async OpenSearch client to improve concurrency ([#125](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/125))
+- [Fix] Close OpenSearch client gracefully ([#126](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/126))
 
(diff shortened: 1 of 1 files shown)


src/mcp_server_opensearch/clusters_information.py
```diff
@@ -116,17 +116,9 @@ async def load_clusters_from_yaml(file_path: str) -> None:
...
-                    result['errors'].append(
-                        f"Error connecting to cluster '{cluster_name}': {error_message}"
-                    )
-                    continue
-                else:
-                    # Add cluster to registry
-                    add_cluster(name=cluster_name, cluster_info=cluster_info)
 
+                # Add cluster to registry without checking connection
+                add_cluster(name=cluster_name, cluster_info=cluster_info)
```
jiapingzeng: Seems that we are not checking cluster connection during startup anymore, instead we store it in registry and connect directly during tool call. I think this is a good idea (since if the mcp server fails to start due to test call failure, the agent wouldn't know the issue anyway) but wanted to check if my understanding is correct?

jiapingzeng (approved):
LGTM, thanks!

PR #134, opened 2025-11-15
Revert "Fixing ListIndexTool when listing all indices (#133)"

rithinpullela:
This reverts commit 00f8cdbbcf17c8a14b557a735ec2731899272592.

### Description
_Describe what this change achieves._

### Issues Resolved

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -16,7 +16,7 @@ async def list_indices(args: ListIndicesArgs) -> json:
     from .client import get_opensearch_client
 
     async with get_opensearch_client(args) as client:
-        response = await client.cat.indices()
+        response = await client.cat.indices(format='json')
         return response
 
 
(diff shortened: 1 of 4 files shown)


src/opensearch/helper.py
```diff
@@ -16,7 +16,7 @@ async def list_indices(args: ListIndicesArgs) -> json:
     from .client import get_opensearch_client
 
     async with get_opensearch_client(args) as client:
-        response = await client.cat.indices()
+        response = await client.cat.indices(format='json')
```
ylwu-amzn, line 19: Cat index result is not json, will this work?

rithinpullela: The cat index result can be a JSON or a list depending on the parameter we send to the OpenSerach py client.

We though this format variable was the root cause of the error. We now realize it is a different issue

PR #135, opened 2025-11-15
Fix AWS auth issues for cat based tools, pin OpenSearchPy to 2.18.0

rithinpullela:
### Description
OpenSearch py Async client causes issues with AWS signing in 3.0 version

When OpenSearch-py of version 3.0 is installed, tools which are based on `client.cat` operation and the search index tool fail because of AWS signing issue.

This issue is only seen now because we are using the async client, this is not seen in sync clients. 

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Added
 
 ### Fixed
-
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -8,7 +8,7 @@ dependencies = [
     "aiohttp>=3.11.18",
     "boto3>=1.38.3",
     "mcp[cli]>=1.9.4",
-    "opensearch-py>=2.8.0",
+    "opensearch-py==2.8.0",
```
ylwu-amzn, line 11: We can't support latest version 3.0.0 ?

rithinpullela: The latest version has a bug which impacts the async client

PR #140, opened 2025-11-19
Convert JSON to CSV for search index tool result

yyfamazon:
### Description
Convert JSON to CSV for search index tool result

Screenshot 2025-11-19 at 15 00 01

### Issues Resolved

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ The following tools are available but disabled by default. To enable them, see t
   - `opensearch_url` (optional): The OpenSearch cluster URL to connect to
   - `index` (required): The name of the index to search in
   - `query` (required): The search query in OpenSearch Query DSL format
+  - `format` (optional): The format of SearchIndexTool response. options are csv and json
 
 - **GetShardsTool**
   - `opensearch_url` (optional): The OpenSearch cluster URL to connect to


src/opensearch/helper.py
```diff
@@ -277,6 +279,60 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+    Returns:
+        str: CSV formatted string of the search results
+    """
+    if not search_results or 'hits' not in search_results:
+        return "No search results to convert"
+    
+    hits = search_results['hits']['hits']
+    if not hits:
+        return "No documents found in search results"
+
```
Hailong-am: what about aggregations queries?

src/opensearch/helper.py
```diff
@@ -277,6 +279,60 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+        row['_score'] = hit.get('_score', '')
+        
+        # Add source fields
+        if '_source' in hit:
+            for field, value in hit['_source'].items():
+                # Convert complex objects to JSON strings
+                if isinstance(value, (dict, list)):
+                    row[field] = json.dumps(value)
+                else:
+                    row[field] = str(value) if value is not None else ''
```
Hailong-am: do we want to expand nested JSON object?

src/opensearch/helper.py
yyfamazon: resolved, thanks!

src/opensearch/helper.py
yyfamazon: resolved, thanks!

src/opensearch/helper.py
```diff
@@ -277,6 +279,182 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+        
+    Returns:
+        str: CSV formatted string of the search results
+    """
+    if not search_results:
+        return "No search results to convert"
+    
+    # Handle aggregations-only queries
+    if 'aggregations' in search_results and ('hits' not in search_results or not search_results['hits']['hits']):
+        return _convert_aggregations_to_csv(search_results['aggregations'])
```
rithinpullela: Some queries might have both aggregations and search hits. Looks like we are either handling the hits or aggregations but not both?

src/opensearch/helper.py
```diff
@@ -277,6 +279,182 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+            _flatten_object(hit['_source'], row)
+        
+        writer.writerow(row)
+    
+    return output.getvalue()
+
+
+def _convert_aggregations_to_csv(aggregations: dict) -> str:
+    """Convert OpenSearch aggregations to CSV format.
+
```
rithinpullela: I am not really sure if aggregations can be converted to a CSV.

Aggregations might not have same format as hits. It can have multiple buckets and deeply nested buckets. Can you please check on this?

I am just wondering if it adds a lot of value to convert aggs to CSV.

tests/opensearch/test_helper.py
```diff
@@ -273,3 +273,178 @@ async def test_get_opensearch_version_error(self, mock_get_client):
...
+        agg_results = {
+            "aggregations": {
+                "status_terms": {
+                    "buckets": [
+                        {"key": "active", "doc_count": 100},
+                        {"key": "inactive", "doc_count": 50}
+                    ]
+                }
+            }
+        }
```
rithinpullela: Can we add more complicated queries here?

Only hits, only aggs, aggs + hits, nested aggs etc?

src/opensearch/helper.py
```diff
@@ -277,6 +279,182 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+        # Add source fields (flattened)
+        if '_source' in hit:
+            _flatten_object(hit['_source'], row)
+        
+        writer.writerow(row)
+    
+    return output.getvalue()
+
+
+def _convert_aggregations_to_csv(aggregations: dict) -> str:
```
rithinpullela: For an aggregation response like this:
```
"aggregations": {
    "brands": {
      "doc_count_error_upper_bound": 0,
      "sum_other_doc_count": 0,
      "buckets": [
        {
          "key": "Apple",
          "doc_count": 70,
          "avg_rating": {
            "value": 4.6
          },
          "review_total": {
            "value": 850
          }
        },
        {
          "key": "Garmin",
          "doc_count": 54,
          "avg_rating": {
            "value": 4.4
          },
          "review_total": {
            "value": 620
          }
        }
      ]
    },
    "rating_buckets": {
      "buckets": [
        {
          "key": 4.0,
          "doc_count": 20
        },
        {
          "key": 4.5,
          "doc_count": 60
        },
        {
          "key": 5.0,
          "doc_count": 44
        }
      ]
    }
  }
```

The CSV might look like this: 
```
brands_avg_rating,brands_doc_count,brands_key,brands_review_total,rating_buckets_doc_count,rating_buckets_key
4.6,70,Apple,,,
4.4,54,Garmin,,,
,,,20,4.0
,,,60,4.5
,,,44,5.0

```

Not really sure if this helps the LLM.

src/opensearch/helper.py
```diff
@@ -277,6 +279,182 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+                _flatten_object(value[0], row, f'{field_name}.')
+                row[field_name] = json.dumps(value)
+            else:
+                # For simple arrays, convert to JSON
+                row[field_name] = json.dumps(value)
+        else:
+            row[field_name] = str(value) if value is not None else ''
+
+
+def _flatten_aggregations(aggs: dict, current_row: dict, rows: list, prefix: str = '') -> None:
```
rithinpullela: I think this func does not handle the below case properly:
```
"aggregations": {
  "brands": {
    "buckets": [
      {
        "key": "Apple",
        "doc_count": 10,
        "avg_rating": { "value": 4.6 }
      }
    ]
  }
}

```

The issue is that rows.append() only happens when a bucket has no nested aggregations. In cases like the example above, we recurse to flatten the metric values, but we never actually append the completed row.

A small change fixes this: if no buckets were seen at this level (meaning we’re at a metric/stat “leaf”), append the populated current_row:
```
if not has_buckets and current_row:
    rows.append(current_row)
```

src/opensearch/helper.py
Hailong-am: I would think it's ok we have one of them, for aggregation queries we can ignore hits as for most of the case size will be 0 and there's no hits. Even it has those hits data wouldn't help LLM as they are partial data.

src/opensearch/helper.py
Hailong-am: the main purpose of csv is to save  tokens, for aggregation it might doesn't have too much  differences.

src/opensearch/helper.py
rithinpullela: Agreed that in most cases we would not have both aggs and hits. And especially when LLMs generate the query, they probably would not generate such queries. But it is not out of the realm of possibilities, and the tool itself should not obscure the actual data returned from OpenSearch right?

PR #151, opened 2025-12-17
Define max response size limit to be applied before data is fetched into memory

arjunkumargiri:
### Description
Define max response size limit to be applied before data is fetched into memory

- Add OPENSEARCH_MAX_RESPONSE_SIZE to environment variables section
- Throw `ResponseSizeExceededError` in case response size breaches configured threshold
- perform chunked reads to prevent memory overflow.
- Document max_response_size parameter for multi-mode cluster configuration
- Add dedicated 'Response Size Limiting' section with detailed explanation

### Testing

Tested by adding MCP server to an agent and specified a small response size to validate that exception is being thrown.

### DIFF
--- a/src/opensearch/client.py
+++ b/src/opensearch/client.py
@@ -32,11 +32,10 @@ OPENSEARCH_SERVICE = 'es'
 OPENSEARCH_SERVERLESS_SERVICE = 'aoss'
 DEFAULT_TIMEOUT = 30
 DEFAULT_SSL_VERIFY = True
-DEFAULT_MAX_RESPONSE_SIZE = 10 * 1024 * 1024  # 10MB default limit
 
 
 # Import custom connection classes and exceptions
-from .connection import BufferedAsyncHttpConnection, ResponseSizeExceededError, OpenSearchClientError
+from .connection import BufferedAsyncHttpConnection, ResponseSizeExceededError, OpenSearchClientError, DEFAULT_MAX_RESPONSE_SIZE
 
 
 class AuthenticationError(OpenSearchClientError):
@@ -388,7 +387,7 @@ def _create_opensearch_client(
         aws_access_key_id: AWS access key ID from headers (optional)
         aws_secret_access_key: AWS secret access key from headers (optional)
         aws_session_token: AWS session token from headers (optional)
-        max_response_size: Maximum response size in bytes (None uses default 10MB)
+        max_response_size: Maximum response size in bytes (None means no limit)
 
     Returns:
         OpenSearch: An initialized OpenSearch client instance
@@ -437,7 +436,10 @@ def _create_opensearch_client(
         'max_response_size': response_size_limit,
     }
     
-    logger.info(f'Configuring OpenSearch client with max_response_size={response_size_limit} bytes')
+    if response_size_limit is not None:
+        logger.info(f'Configuring OpenSearch client with max_response_size={response_size_limit} bytes')
+    else:
+        logger.info('Configuring OpenSearch client with no response size limit')
 
     # Create boto3 session
     try:
--- a/src/opensearch/connection.py
+++ b/src/opensearch/connection.py
@@ -15,7 +15,7 @@ from opensearchpy import AsyncHttpConnection
 logger = logging.getLogger(__name__)
 
 # Constants
-DEFAULT_MAX_RESPONSE_SIZE = 10 * 1024 * 1024  # 10MB default limit
+DEFAULT_MAX_RESPONSE_SIZE = None  # No limit by default - only enforce when explicitly set
 
 
 # Base exception class (to avoid circular imports)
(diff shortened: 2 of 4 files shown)


src/opensearch/connection.py
```diff
@@ -0,0 +1,247 @@
...
+                        )
+                        logger.error(
+                            f'Response size exceeded limit during streaming: '
+                            f'{total_size + len(chunk)} > {self.max_response_size} bytes'
+                        )
+                        raise ResponseSizeExceededError(
+                            f"Response size exceeded limit of {self.max_response_size} bytes. "
+                            f"Stopped reading at {total_size} bytes to prevent memory exhaustion. "
+                            f"Consider increasing max_response_size or refining your query to return less data."
+                        )
```
Hailong-am, line 165: Instead of raise an error, do you think we could have an option to return truncated data for non `http_compress` case?

Hailong-am: also shall we consider the case with http_compress. gzip usually have `**JSON/XML**: 80-95% (5x-20x smaller)` https://lemire.me/blog/2021/06/30/compressing-json-gzip-vs-zstd/

arjunkumargiri: LLM would then need to know to decompress the data. Otherwise we should let the LLM know that the data would be compressed. 

Irrespective I believe larger data will fill up the context quickly, so it might be better to let the LLM know that the response data is too large so that the LLM could try a different mechanism. For example, if the LLM is trying to run a search query that returns all the documents, it might be better to let the LLM know the response is large so that if could try to aggregate the required information instead.

Hailong-am: for http_compress case, shouldn't we use a lower value? if we keep using 10M as default, after decompress, it might become to 60MB, that's too large for LLM.

arjunkumargiri: > In that case, what do you think about returning the truncated response?

My concern with truncated response is that the LLM might take a decision with a subset of data without considering complete context which could result in incorrect response. Even if we include a "Response is truncated", I am not sure if LLM will ignore that information since there is also lot of other information which is returned from the actual reponse. 

>  think throwing exception might waste resources since the request is already made and we are fetching the response.

Amount of resource wasted to query opensearch should be much smaller when compared to LLM resource consumed by passing a in-complete data.

> for http_compress case, shouldn't we use a lower value? if we keep using 10M as default, after decompress, it might become to 60MB, that's too large for LLM.

Agreed, currently using 10 MB as a safe limit so that existing usecases is not broken. User will need to set a explicit lower value in settings to efficiently use this feature.

src/opensearch/client.py
```diff
@@ -32,13 +32,11 @@
 OPENSEARCH_SERVERLESS_SERVICE = 'aoss'
 DEFAULT_TIMEOUT = 30
 DEFAULT_SSL_VERIFY = True
+DEFAULT_MAX_RESPONSE_SIZE = 10 * 1024 * 1024  # 10MB default limit
```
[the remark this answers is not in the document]
arjunkumargiri: Ack, I was going for a safe upper limit, but I agree that by default we don't have to define a limit.

src/opensearch/connection.py
arjunkumargiri: few options available for LLM:
* Add more filters so that subset of data is fetched.
* Perform aggregations on server side rather than pulling complete set of documents to client side and then performing aggregation
* Define smaller size limit and then perform multiple calls (Although this will still impact context window.)

src/opensearch/client.py
```diff
@@ -274,6 +285,20 @@ def _initialize_client_multi_mode(cluster_info: ClusterInfo) -> AsyncOpenSearch:
...
+        # Get max response size from cluster config, fallback to environment variable
+        max_response_size = cluster_info.max_response_size
+        if max_response_size is None:
+            max_response_size_str = os.getenv('OPENSEARCH_MAX_RESPONSE_SIZE', '').strip()
+            if max_response_size_str:
+                try:
+                    max_response_size = int(max_response_size_str)
+                    if max_response_size <= 0:
+                        logger.warning(f'Invalid OPENSEARCH_MAX_RESPONSE_SIZE value {max_response_size}, using default')
+                        max_response_size = None
```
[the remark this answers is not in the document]
mingshl: for opensearch setting, the defalut is None

USER_GUIDE.md
```diff
@@ -441,6 +441,12 @@ python -m mcp_server_opensearch --mode multi
 |----------|----------|---------|-------------|
 | `OPENSEARCH_SSL_VERIFY` | No | `"true"` | Control SSL certificate verification (`"true"` or `"false"`) |
 
+### Response Size Control Variables
+
+| Variable | Required | Default | Description |
+|----------|----------|---------|-------------|
+| `OPENSEARCH_MAX_RESPONSE_SIZE` | No | `10485760` (10MB) | Maximum response size in bytes to prevent memory exhaustion from large responses |
```
mingshl, line 448: can we leverage this opensearch setting for the request size as well? 
https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/network-settings/
```
http.max_content_length (Static, byte unit): Sets the maximum content length allowed for HTTP requests. Requests exceeding this limit will be rejected. This setting helps prevent memory issues caused by extremely large requests. Default is 100mb.```

arjunkumargiri: We need a mechanism to control the response value on client side, provided setting might not help for memory usage issue.

PR #157, opened 2025-12-29
Fix _fallback_perform_request using wrong url

Hailong-am:
### Description
Fix _fallback_perform_request using wrong full url, it should use original url.

If a request failed, the url will changed to a full http endpoint at line 103

https://github.com/opensearch-project/opensearch-mcp-server-py/blob/0105d4a75d49aff55b52e498e7ac63f1cd00210e/src/opensearch/connection.py#L100-L103

and _fallback_perform_request requires the original url as parameter, otherwise the url will become `https://opensearch:443https://opensearch:9210/index/_search`

```
Error searching index: Port could not be cast to integer value as '443https:
```

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Support to define max response size from opensearch cluster for better memory management
 
 ### Fixed
+- Fix _fallback_perform_request using wrong url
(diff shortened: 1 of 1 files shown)


Hailong-am:
@arjunkumargiri would you mind to review it

arjunkumargiri (approved):

nathaliellenaa (approved):

PR #171, opened 2026-02-24
Add search configuration tools 

wrigleyDan:
Add search configuration tools for Search Relevance Workbench interactions.

### Description
This PR creates three new tools to handle search configurations (part of the Search Relevance Workbench). As a side effect it upgrades the OpenSearch Python client to support them.
This PR is a first attempt to include functionality necessary for future agents in OpenSearch following conversations with @mingshl and @heemin32 to integrate the agent for agentic relevance tuning into OpenSearch. It is part of the larger effort under the RFC for [remote agent support](https://github.com/opensearch-project/OpenSearch/issues/20602).

**Changes**

Dependency upgrade (`pyproject.toml`):
- Upgrades opensearch-py from 2.8.0 to 3.1.0, which adds native plugin support for the Search Relevance API via client.plugins.search_relevance.

Three new MCP tools following the standard 3-layer architecture (args model → helper → tool function → registry):
- `CreateSearchConfigurationTool` — creates a search configuration via `PUT /_plugins/_search_relevance/search_configurations`
- `GetSearchConfigurationTool` — retrieves a search configuration by ID via `GET /_plugins/_search_relevance/search_configurations/{id}`
- `DeleteSearchConfigurationTool` — deletes a search configuration by ID via `DELETE /_plugins/_search_relevance/search_configurations/{id}`

All three tools require OpenSearch cluster version ≥ 3.1.0 (enforced via `min_version` in the tool registry).

New search_relevance tool category (`tool_filter.py`):
- The three tools are grouped under a new built-in `search_relevance` category.
- The category is not enabled by default. Users opt in via `OPENSEARCH_ENABLED_CATEGORIES=search_relevance` or by adding `enabled_categories: [search_relevance]` or explicitly adding individual tools to their config file.

Tests:
- Unit tests for all three helper functions, tool functions (success and error paths), and the new category filter behaviour.

### Issues Resolved
n/a

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -55,6 +55,9 @@ async def search_index(args: SearchIndexArgs) -> json:
     from .client import get_opensearch_client
     from tools.tools import TOOL_REGISTRY
 
+    if isinstance(args.query, str):
+        validate_json_string(args.query)
+
     async with get_opensearch_client(args) as client:
         query = normalize_scientific_notation(args.query)
         
@@ -507,6 +510,8 @@ async def create_search_configuration(args: CreateSearchConfigurationArgs) -> js
     """
     from .client import get_opensearch_client
 
+    validate_json_string(args.query)
+
     async with get_opensearch_client(args) as client:
         body = {
             'name': args.name,
@@ -551,6 +556,27 @@ async def delete_search_configuration(args: DeleteSearchConfigurationArgs) -> js
             search_configuration_id=args.search_configuration_id
         )
         return response
+
+
+def validate_json_string(value: str) -> None:
+    """Validate that a string is valid JSON, raising ValueError with a concise message if not.
+
+    Intended to be called early (before any API call) so the LLM receives a small,
+    precise error rather than a verbose OpenSearch response.
+
+    Args:
+        value: The string to validate as JSON.
+
+    Raises:
+        ValueError: If the string is not valid JSON. The message includes the parse
+            error description, line, and column so the problem is immediately obvious.
+    """
+    try:
+        json.loads(value)
+    except json.JSONDecodeError as e:
+        raise ValueError(
+            f"query is not valid JSON: {e.msg} (line {e.lineno}, col {e.colno})"
+        ) from e
 
 
 def normalize_scientific_notation(body):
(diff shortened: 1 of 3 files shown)


rithin-pullela-aws:
Thanks for the PR @wrigleyDan, looks like we upgraded the OpenSearch-py version to 3.1, is that a requirement for the new tools?

The reason we locked it to 2.8 was because we were seeing some bugs with Async classes in the "then latest" 3.0
Before we upgrade back to latest version 3.1. I would like to test out Auth behavior in different configurations.

src/opensearch/helper.py
```diff
@@ -496,6 +496,63 @@ def _convert_value(v):
...
+        json: OpenSearch response with the created configuration ID
+    """
+    from .client import get_opensearch_client
+
+    async with get_opensearch_client(args) as client:
+        body = {
+            'name': args.name,
+            'index': args.index,
+            'query': args.query,  # must remain a JSON string, not a dict
+        }
```
rithinpullela, line 520: nit: sometimes LLMs find it hard to create a JSON string and might mess up escaping.

Can we consider validating if the query argument from LLM is a valid JSON String? If not we can fail early with a small error message (low tokens, showcasing the exact problem) rather than waiting for OpenSearch to throw an error?

rithin-pullela-aws:
> Thanks for the PR @wrigleyDan, ~looks like we upgraded the OpenSearch-py version to 3.1, is that a requirement for the new tools?~
> 
> ~The reason we locked it to 2.8 was because we were seeing some bugs with Async classes in the "then latest" 3.0 Before we upgrade back to latest version 3.1. I would like to test out Auth behavior in different configurations.~
> 
> Just saw your PR description mentioning it. Let me test various auths to make sure we don't run into any bugs

Update:
I tested with version 3.1 and all the Auths work as expected, the upgrade is not a blocker for this PR

src/opensearch/helper.py
wrigleyDan: Great idea, that makes sense! Added a validation function.

src/opensearch/helper.py
```diff
@@ -55,6 +55,9 @@ async def search_index(args: SearchIndexArgs) -> json:
     from .client import get_opensearch_client
     from tools.tools import TOOL_REGISTRY
 
+    if isinstance(args.query, str):
+        validate_json_string(args.query)
+
```
rithinpullela, line 60: Thanks for adding it here as well!

PR #178, opened 2026-02-27
Structured JSON Logging for Monitoring and Metrics

rithinpullela:
### Summary

Adds opt-in structured JSON logging (`--log-format json`) that emits machine-parseable events for tool executions, errors, HTTP requests, connection failures, and memory usage. Users who forward logs to any monitoring platform (CloudWatch, Datadog, Splunk, Grafana Loki, ELK, etc.) can build metric filters, alarms, and dashboards directly from these structured fields.

Default behavior (`--log-format text`) is unchanged.

Full design details: https://github.com/opensearch-project/opensearch-mcp-server-py/issues/177

### Changes

**New files:**
- `logging_config.py` — `JsonFormatter`, `configure_logging()`, periodic `memory_monitor()`
- `tool_executor.py` — Shared `execute_tool()` with timing, error detection, structured logging
- `tool_logging.py` — `log_tool_error()` helper extracting `status_code`, `exception_type`, `root_cause` from opensearchpy exceptions

**Modified files:**
- `__init__.py` — Added `--log-format` CLI flag
- `stdio_server.py` / `streaming_server.py` — Delegate `call_tool()` to shared executor
- `connection.py` — Emit `opensearch_request` events per HTTP call
- `client.py` — Emit `datasource_connection` events on auth failures
- `tools.py`, `tool_generator.py`, `generic_api_tool.py`, `skills_tools.py` — Route all errors through `log_tool_error()`

### Event Types

| Event | When | Key Fields |
|-------|------|------------|
| `tool_execution` | Every tool call | `tool_name`, `status`, `duration_ms` |
| `tool_error` | Tool failure | `tool_name`, `status_code`, `exception_type`, `root_cause` |
| `opensearch_request` | Every HTTP request | `http_method`, `endpoint`, `status_code`, `duration_ms` |
| `datasource_connection` | Auth/connection failure | `auth_method`, `datasource_type`, `opensearch_url` |
| `memory_snapshot` | Every 60s | `memory_rss_mb`, `pid` |

### Testing

- All unit tests pass (`pytest tests/`) — 11 tests for `tool_logging`, 8 for `tool_executor`, 12+ for `logging_config`
- Integration tested with 8 scenarios against a live OpenSearch cluster covering success, 404, 403, DNS failure, and SigV4 retry patterns
- Backward compatible — `--log-format text` output is identical to previous behavior

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes https://github.com/opensearch-project/opensearch-mcp-server-py/issues/177

### DIFF
--- a/src/tools/generic_api_tool.py
+++ b/src/tools/generic_api_tool.py
@@ -88,12 +88,11 @@ async def generic_opensearch_api_tool(args: GenericOpenSearchApiArgs) -> list[di
         valid_methods = ['GET', 'POST', 'PUT', 'DELETE', 'HEAD', 'PATCH']
         method = args.method.upper()
         if method not in valid_methods:
-            return [
-                {
-                    'type': 'text',
-                    'text': f'Error: Invalid HTTP method "{args.method}". Valid methods are: {", ".join(valid_methods)}',
-                }
-            ]
+            return log_tool_error(
+                'GenericOpenSearchApiTool',
+                ValueError(f'Invalid HTTP method "{args.method}". Valid methods are: {", ".join(valid_methods)}'),
+                'validating request',
+            )
 
         # Check if write operations are allowed using the global setting
         # Import here to avoid circular import (tool_filter -> tools -> generic_api_tool -> tool_filter)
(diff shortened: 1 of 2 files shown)


src/mcp_server_opensearch/logging_config.py
```diff
@@ -0,0 +1,158 @@
...
+        import resource
+
+        return round(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / (1024 * 1024), 2)
+    except ImportError:
+        pass
+
+    return -1.0
+
+
+async def memory_monitor(interval_seconds: int = 60) -> None:
```
nathaliellenaa, line 110: Should we make this intervals configurable?

src/mcp_server_opensearch/tool_executor.py
```diff
@@ -0,0 +1,122 @@
...
+        # (or prefix error text with "Error") for this detection to work.
+        if result and len(result) > 0:
+            text = ''
+            if isinstance(result[0], dict):
+                text = result[0].get('text', '')
+            elif hasattr(result[0], 'text'):
+                text = result[0].text
+
+            if text.startswith('Error'):
+                status = 'error'
```
nathaliellenaa: This relies on an implicit contract that error messages start with "Error", which seems brittle since any tool returning text that happens to start with "Error" (e.g., "Error codes explained:") would be incorrectly flagged as a failure. One suggestion could be returning a structured response with an explicit status indicator. For example:
```
# Tools return a structured response:
return [{'type': 'text', 'text': '...', 'is_error': True}]

# Then in execute_tool, check the status explicitly:
if result and len(result) > 0:
    if isinstance(result[0], dict) and result[0].get('is_error'):
        status = 'error'
```
This way, error detection is based on an explicit flag rather than string pattern matching, making it more reliable and less prone to false positives.

nathaliellenaa (commented):
Overall changes looks good @rithin-pullela-aws! Added some high level comments

src/mcp_server_opensearch/logging_config.py
rithinpullela: Makes sense, we can have an env variable for this

rithinpullela: Added an env variable called `OPENSEARCH_MEMORY_MONITOR_INTERVAL ` and we use default of 60 sec when the variable is not set

src/mcp_server_opensearch/tool_executor.py
rithinpullela: This is a great point! Addressed this comment in my latest commits

PR #187, opened 2026-03-03
add tools to manage query sets in the Search Relevance Workbench

wrigleyDan:
Add query set tools for Search Relevance Workbench interactions.

### Description

This PR creates new tools to handle query sets (part of the Search Relevance Workbench)

This PR extends the already added search configuration handling tools to include functionality necessary for agentic relevance tuning into OpenSearch. It is part of the larger effort under the RFC for [remote agent support](https://github.com/opensearch-project/OpenSearch/issues/20602).

Changes

Four new MCP tools following the standard 3-layer architecture (args model → helper → tool function → registry):

`CreateQuerySetTool` — creates a query set via `PUT /_plugins/_search_relevance/query_sets`
`SampleQuerySetTool` — samples a query set from UBI data via `POST /_plugins/_search_relevance/query_sets`
`GetQuerySetTool` — retrieves a query set by ID via `GET /_plugins/_search_relevance/query_sets/{id}`
`DeleteQuerySetTool` — deletes a query set by ID via `DELETE /_plugins/_search_relevance/query_sets/{id}`
All four tools require OpenSearch cluster version ≥ 3.1.0 (enforced via `min_version` in the tool registry).
All tools belong to the category `search_relevance`.

Tests:

Unit tests for all three helper functions, tool functions (success and error paths), and the new category filter behaviour.

### Issues Resolved

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -368,8 +368,8 @@ async def sample_query_set(args: SampleQuerySetArgs) -> json:
 
     body = {
         'name': args.name,
-        'description': args.description or f'Top {args.query_set_size} most frequent queries',
-        'sampling': 'topn',
+        'description': args.description or f'Query set: {args.name} ({args.sampling}, size={args.query_set_size})',
+        'sampling': args.sampling,
         'querySetSize': args.query_set_size,
     }
 
--- a/src/tools/tool_params.py
+++ b/src/tools/tool_params.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from pydantic import BaseModel, Field
-from typing import Any, Optional, Type, TypeVar, Dict
+from typing import Any, Literal, Optional, Type, TypeVar, Dict
 from mcp_server_opensearch.global_state import get_mode
 
 T = TypeVar('T', bound=BaseModel)
@@ -338,6 +338,15 @@ class SampleQuerySetArgs(baseToolArgs):
     query_set_size: int = Field(
         default=20, description='Number of top queries to sample (default: 20)', ge=1
     )
+    sampling: Literal['topn', 'random', 'pptss', 'all'] = Field(
+        default='topn',
+        description=(
+            'Sampling method: "topn" (most frequent N queries), '
+            '"random" (random sample), '
+            '"pptss" (probability-proportional-to-size sampling), '
+            '"all" (all queries)'
+        ),
+    )
     description: str = Field(default='', description='Optional description of the query set')
 
     class Config:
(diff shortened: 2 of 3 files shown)


src/opensearch/helper.py
```diff
@@ -297,6 +297,115 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+    Args:
+        args: DeleteQuerySetArgs containing the query_set_id
+
+    Returns:
+        json: Result of the deletion operation
+    """
+    from .client import get_opensearch_client
+
+    async with get_opensearch_client(args) as client:
+        response = await client.transport.perform_request(
```
epugh: Could we have used the built in method https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/plugins/search_relevance.py#L100 ?   

Here is a list of methods that we could use through the python client...  

Image

epugh (commented):
One comment about if we want to leverage the OpenAPI annotated methods versus writing our own..  The rest looks reaosnable.

src/tools/tools.py
```diff
@@ -558,6 +566,78 @@ async def delete_search_configuration_tool(args: DeleteSearchConfigurationArgs)
...
+    Returns:
+        list[dict]: Query set details in MCP format
+    """
+    try:
+        await check_tool_compatibility('GetQuerySetTool', args)
+        result = await get_query_set(args)
+        formatted_result = json.dumps(result, separators=(',', ':'))
+        return [{'type': 'text', 'text': f'Query set {args.query_set_id}:\n{formatted_result}'}]
+    except Exception as e:
+        return [{'type': 'text', 'text': f'Error retrieving query set: {str(e)}'}]
```
nathaliellenaa: Error handling should use `log_tool_error` for consistency with the other tools

src/opensearch/helper.py
```diff
@@ -297,6 +297,115 @@ async def get_nodes_info(args: GetNodesArgs) -> json:
...
+
+    Returns:
+        json: Result of the sampling operation with the created query set ID
+    """
+    from .client import get_opensearch_client
+
+    body = {
+        'name': args.name,
+        'description': args.description or f'Top {args.query_set_size} most frequent queries',
+        'sampling': 'topn',
```
nathaliellenaa: Is the `topn` sampling strategy intentionally hardcoded here? Should we make this configurable via SampleQuerySetArgs?

src/opensearch/helper.py
wrigleyDan: Thanks, I followed the wrong pattern here. It makes total sense to use the methods we have in the Python client. I changed that.

src/tools/tools.py
wrigleyDan: Thanks, changed.

src/opensearch/helper.py
wrigleyDan: It should absolutely be configurable, good catch, I made the changes. Thank you 👍

PR #194, opened 2026-03-10
Tune searchIndexTool description

Hailong-am:
### Description
Tune searchIndexTool description to let's the tool get mapping first before construct DSL query

### Issues Resolved


Hailong-am:
> LGTM! Thanks @Hailong-am.
> 
> On a side note @Hailong-am do you think adding DSL guidance "skills" (essentially a .md file wrapped in a tool) would help here?
> 
> I have seen the LLMs finding it hard to write DSL queries. And they might need guidance about DSL more often than not. We can perhaps expose a bunch of skills like: basic_query_rules, aggregation_rules, neural_rules etc and the agent can read the SOP/guide once to understand DSL rules when it is unable to generate DSL on its own

@rithin-pullela-aws  Thanks a good suggestion. I have considered this as well and one concern is that LLM already have those kind of knowledge, not sure how LLM will use these skills and enforce the rules.

PR #202, opened 2026-03-24
Cache OpenSearch version to avoid redundant HTTP calls per tool invocation

### LINKED ISSUE #185: [BUG] Tools take a long time to respond
### What is the bug?
In the OpenSearch MCP server we make unnecessary calls to OpenSearch cluster for version validation and other auxiliary tasks. This increases the latency of the tools. 

### How can one reproduce the bug?
Make a cat indices API call and the use OpenSearch MCP server list indices tool.
The latency from MCP tool is ~2x 

### What is the expected behavior?
The latency from the tools should be very close to raw API call

### Solution
Do not make multiple HTTP calls when unnecessary

ShivaniNR:
### Description
Cache the OpenSearch cluster version after the first successful retrieval to eliminate the redundant HTTP call made by `get_opensearch_version()` on every tool invocation. Previously, each tool call made 2 HTTP round-trips (version check + actual operation); now, subsequent calls return the cached version.

- Add `_version_cache` module-level dict in `helper.py` keyed by cluster name
- Add `clear_version_cache()` for cache reset
- Add autouse conftest fixture to prevent cache leaking between tests 
- Add 4 tests covering cache hit, per-cluster isolation, failure retry, and cache clearing

### Issues Resolved
Closes #185 

### DIFF
--- a/src/opensearch/helper.py
+++ b/src/opensearch/helper.py
@@ -686,18 +686,38 @@ def _flatten_object(obj: dict, row: dict, prefix: str = '') -> None:
             row[field_name] = str(value) if value is not None else ''
 
 
+_version_cache: dict[str, Version] = {}
+
+
+def clear_version_cache():
+    """Clear the cached OpenSearch version(s).
+
+    Useful for testing or when cluster version changes.
+    """
+    _version_cache.clear()
+
+
 async def get_opensearch_version(args: baseToolArgs) -> Version:
     """Get the version of OpenSearch cluster.
+
+    Returns the cached version if available, otherwise fetches it from the
+    cluster and caches the result. Failed lookups (None) are not cached.
 
     Returns:
         Version: The version of OpenSearch cluster (SemVer style)
     """
     from .client import get_opensearch_client
 
+    cache_key = args.opensearch_cluster_name or ''
+    if cache_key in _version_cache:
+        return _version_cache[cache_key]
+
     try:
         async with get_opensearch_client(args) as client:
             response = await client.info()
-            return Version.parse(response['version']['number'])
+            version = Version.parse(response['version']['number'])
+            _version_cache[cache_key] = version
+            return version
     except Exception as e:
         logger.error(f'Error getting OpenSearch version: {e}')
         return None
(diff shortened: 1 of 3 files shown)


src/opensearch/helper.py
```diff
@@ -686,18 +686,38 @@ def _flatten_object(obj: dict, row: dict, prefix: str = '') -> None:
...
+    """
+    _version_cache.clear()
+
+
 async def get_opensearch_version(args: baseToolArgs) -> Version:
     """Get the version of OpenSearch cluster.
 
+    Returns the cached version if available, otherwise fetches it from the
+    cluster and caches the result. Failed lookups (None) are not cached.
+
```
rithinpullela, line 705: Great implementation @ShivaniNR! 

Can we add a ttl for this cache? This could be helpful in a very small edge case where te cluster gets upgraded when the MCP server is still running. 

The code can look something like this, let me know what you think about this:
```
 import time

  _version_cache: dict[str, tuple[Version, float]] = {}
  _VERSION_CACHE_TTL = int(os.getenv('OPENSEARCH_VERSION_CACHE_TTL', '300'))


  def clear_version_cache():
      """Clear the cached OpenSearch version(s).

      Useful for testing or when cluster version changes.
      """
      _version_cache.clear()


  async def get_opensearch_version(args: baseToolArgs) -> Version:
      """Get the version of OpenSearch cluster.

      Returns the cached version if available and not expired, otherwise
      fetches it from the cluster and caches the result. Failed lookups
      (None) are not cached.

      Returns:
          Version: The version of OpenSearch cluster (SemVer style)
      """

      cache_key = args.opensearch_cluster_name or ''
      if cache_key in _version_cache:
          version, cached_at = _version_cache[cache_key]
          if time.time() - cached_at < _VERSION_CACHE_TTL:
              return version

      try:
          async with get_opensearch_client(args) as client:
              response = await client.info()
              version = Version.parse(response['version']['number'])
              _version_cache[cache_key] = (version, time.time())
              return version
      except Exception as e:
          logger.error(f'Error getting OpenSearch version: {e}')
          return None

```

rithinpullela (commented):
Thanks for the PR @ShivaniNR! 
Added a small comment, can you also add changelog for this PR

PR #204, opened 2026-03-28
added mtls support for opensearch clusters. supports both single and multi-mode

### LINKED ISSUE #168: MTLS support for the opensearch client
## Problem Overview
**opensearch-mcp-server-py** cannot connect to OpenSearch clusters where **HTTP TLS client authentication** (clientauth_mode: REQUIRE) is enabled.

Although opensearch-py (including AsyncOpenSearch) supports mTLS via client_cert, client_key, or a custom ssl_context, the MCP server does not expose or wire these options when constructing the async client. As a result, connections fail during the TLS handshake with:

```bash
ssl/tls alert bad certificate (alert 42)
```

This occurs even when:
- the server CA is trusted
- valid client certificates and keys are provided
- the same certificates work with openssl client

At present, there is no configuration-only workaround; users must either disable mTLS, terminate mTLS outside the MCP server (proxy/sidecar), or fork and patch the server to pass client TLS settings.

## Expected behavior:
The MCP server should allow configuring mTLS for the async OpenSearch client, e.g. via environment variables:
OPENSEARCH_CLIENT_CERT
OPENSEARCH_CLIENT_KEY
(optionally) custom SSLContext
and pass these to AsyncOpenSearch.

kabirrajsingh:
## Summary

  This PR adds outbound mutual TLS (mTLS) support to the Python OpenSearch MCP server so it can connect to OpenSearch clusters that require client certificates while preserving old behaviour

  It introduces mTLS configuration for both:
  - single mode via environment variables
  - multi mode via YAML cluster config

  It also adds:
  - validation for mTLS certificate/key path combinations
  - updated docs and example config
  - unit test coverage for env parsing, YAML parsing, and client construction

  Closes #168 (https://github.com/opensearch-project/opensearch-mcp-server-py/issues/168)

  ## Motivation

  The current server supports basic auth, AWS auth, header auth, and no-auth, but does not support OpenSearch clusters that require client certificate authentication.

  This blocks deployment scenarios where:
  - the MCP server runs inside Kubernetes
  - OpenSearch is accessible only over mTLS
  - the MCP endpoint needs to be exposed separately for remote clients

  This PR adds the missing transport-layer TLS configuration without changing the MCP tool interface.

  ## What Changed

  ### OpenSearch client
  - Added support for:
    - `OPENSEARCH_CA_CERT_PATH`
    - `OPENSEARCH_CLIENT_CERT_PATH`
    - `OPENSEARCH_CLIENT_KEY_PATH`
  - Threaded these through single-mode and multi-mode client initialization
  - Passed the resulting TLS kwargs into `AsyncOpenSearch`
  - Added validation:
    - client cert and client key must be provided together
    - configured TLS files must exist

  ### Multi-cluster config
  - Extended `ClusterInfo` with:
    - `opensearch_ca_cert_path`
    - `opensearch_client_cert_path`
    - `opensearch_client_key_path`
  - Added YAML loading support for these fields

  ## Testing

  Ran:
  ```bash
  uv run pytest tests/opensearch/test_client.py tests/opensearch/test_response_size_limiting.py tests/mcp_server_opensearch/test_clusters_information.py
  uv run pytest tests/mcp_server_opensearch/test_streaming_server.py

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -277,6 +277,18 @@ export OPENSEARCH_USERNAME="<your_opensearch_domain_username>"
 export OPENSEARCH_PASSWORD="<your_opensearch_domain_password>"
 ```
 
+#### Mutual TLS (mTLS)
+```bash
+export OPENSEARCH_URL="<your_opensearch_domain_url>"
+export OPENSEARCH_USERNAME="<your_opensearch_domain_username>"
+export OPENSEARCH_PASSWORD="<your_opensearch_domain_password>"
+export OPENSEARCH_CA_CERT_PATH="/path/to/ca.crt"
+export OPENSEARCH_CLIENT_CERT_PATH="/path/to/tls.crt"
+export OPENSEARCH_CLIENT_KEY_PATH="/path/to/tls.key"
+```
+
+`OPENSEARCH_CLIENT_CERT_PATH` and `OPENSEARCH_CLIENT_KEY_PATH` must be provided together. These settings add TLS client certificates to the existing authentication flow, so you can use mTLS alongside basic auth, IAM, or no-auth clusters.
+
 #### AWS Credentials Authentication
 ```bash
 export OPENSEARCH_URL="<your_opensearch_domain_url>"
@@ -314,6 +326,15 @@ clusters:
     opensearch_url: "http://localhost:9200"
     opensearch_username: "admin"
     opensearch_password: "your_password_here"
+
+  # Mutual TLS (mTLS)
+  mtls-cluster:
+    opensearch_url: "https://private-opensearch.example.com:9200"
+    opensearch_username: "admin"
+    opensearch_password: "your_password_here"
+    opensearch_ca_cert_path: "/var/run/opensearch-mtls/ca.crt"
+    opensearch_client_cert_path: "/var/run/opensearch-mtls/tls.crt"
+    opensearch_client_key_path: "/var/run/opensearch-mtls/tls.key"
 
   # AWS Credentials Authentication
   remote-cluster:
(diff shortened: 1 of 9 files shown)


kabirrajsingh:
can this please be reviewed @nathaliellenaa @rithinpullela @ylwu-amzn

USER_GUIDE.md
```diff
@@ -417,6 +452,21 @@ python -m mcp_server_opensearch --mode multi --config config.yml --profile my-aw
...
+1. Build the image with the included [Dockerfile](./Dockerfile).
+2. Create a secret containing `ca.crt`, `tls.crt`, and `tls.key`.
+3. Mount that secret into the pod and set:
+   - `OPENSEARCH_CA_CERT_PATH`
+   - `OPENSEARCH_CLIENT_CERT_PATH`
+   - `OPENSEARCH_CLIENT_KEY_PATH`
+4. Expose the pod through a `Service`.
+5. Publish `/mcp` and `/health` through an `Ingress`.
+
+See [deploy/k8s/mcp-server.example.yaml](./deploy/k8s/mcp-server.example.yaml) for a complete example manifest.
```
rithinpullela: Looks like these files do not exist, can you take a look?

src/opensearch/client.py
```diff
@@ -621,6 +650,67 @@ def _create_opensearch_client(
...
+    if not normalized_path:
+        return None
+
+    return normalized_path
+
+
+def _validate_tls_file_path(path: str, description: str) -> str:
+    """Validate that a configured TLS file path exists."""
+    if not os.path.isfile(path):
+        raise ConfigurationError(f'{description} file does not exist or is not a file: {path}')
```
rithinpullela: Can we also check if we have read permissions? `os.access(path, os.R_OK)`

rithinpullela (commented):
Thanks for the PR @kabirrajsingh! Overall looks good, added 2 small comments. Also can you please add changelog?

USER_GUIDE.md
kabirrajsingh: Removed the line. Had added them to provide examples to use them in kubernetes clusters but did not commit them since I though we can have one seperate PR with all examples using different kind of auth so it doesn't confuse people.

src/opensearch/client.py
kabirrajsingh: added the check

kabirrajsingh:
added changelog and made the changes for those 2 comments

kabirrajsingh:
@rithinpullela can you please have a look

rithinpullela (approved):
Thanks for the PR @kabirrajsingh@! LGTM

PR #205, opened 2026-03-30
Add readonly-hint annotation to reaodnly MCP tools

bbartels:
`list_tools()` in both stdio and streaming servers constructed `Tool` objects with no `annotations`, so MCP clients never received `readOnlyHint` signals even though the registry already encodes HTTP method metadata sufficient to derive read-only status.

## Changes

- **`src/tools/utils.py`** — New `is_read_only_tool(tool_info)` helper: parses `http_methods` (string or list) and returns `True` iff the method set is exclusively `{'GET'}`
- **`stdio_server.py` / `streaming_server.py`** — Both `list_tools()` handlers now pass `annotations=ToolAnnotations(readOnlyHint=is_read_only_tool(tool_info))` when constructing each `Tool`
- **`test_streaming_server.py` / `test_stdio_server.py`** — New `test_list_tools_readonly_hint` in each: exercises the actual registered handler via `request_handlers[ListToolsRequest]` and asserts `readOnlyHint=True` for a GET-only tool and `readOnlyHint=False` for a POST tool

## Example

```python
# Before — annotations absent, clients assume tools may have side effects
Tool(name="ListIndexTool", description="...", inputSchema={...})

# After — GET-only tools advertise readOnlyHint=True
Tool(
    name="ListIndexTool",
    description="...",
    inputSchema={...},
    annotations=ToolAnnotations(readOnlyHint=True),   # derived from http_methods='GET'
)
```

> [!WARNING]
>
> <details>
> <summary>Firewall rules blocked me from connecting to one or more addresses (expand for details)</summary>
>
> #### I tried to connect to the following addresses, but was blocked by firewall rules:
>
> - `test-opensearch-domain.com`
>   - Triggering command: `/usr/bin/python python -m pytest tests/ -v --ignore=tests/tools/test_judgment_tool_evals.py -x` (dns block)
>   - Triggering command: `/usr/bin/python python -m pytest tests/ -v --ignore=tests/tools/test_judgment_tool_evals.py` (dns block)
>
> If you need me to access, download, or install something from one of these locations, you can either:
>
> - Configure [Actions setup steps](https://gh.io/copilot/actions-setup-steps) to set up my environment, which run before the firewall is enabled
> - Add the appropriate URLs or hosts to the custom allowlist in this repository's [Copilot coding agent settings](https://github.com/bbartels/opensearch-mcp-server-py/settings/copilot/coding_agent) (admins only)
>
> </details>

<details>

<summary>Original prompt</summary>

Add MCP readOnlyHint publishing for read-only tools in the repository bbartels/opensearch-mcp-server-py.

Context:
- The server currently publishes MCP tools from both stdio and streaming servers by constructing mcp.types.Tool objects with only name, description, and inputSchema.
- The MCP readOnlyHint annotation does not appear to be set anywhere today.
- Tool metadata already includes http_methods in the registry, and read-only tools are effectively identifiable from that metadata.
- The user wants a PR raised against their fork.

Requested change:
1. Update tool definitions/publication so that tools that are read-only publish the MCP readOnlyHint annotation.
2. Ensure the hint is included in the Tool objects returned by list_tools() in both server implementations.
3. Use the existing tool metadata to determine which tools are read-only, unless a better minimal design is already natural in the codebase.
4. Preserve existing behavior for name, description, inputSchema, filtering, and tool execution.
5. Add or update tests to cover the new behavior, including at least one read-only tool and one non-read-only tool if applicable.
6. Create a pull request in bbartels/opensearch-mcp-server-py.

Implementation guidance:
- Prefer a minimal, maintainable patch.
- If the MCP SDK Tool model supports annotations directly, use that.
- If tool metadata needs to be extended, do so consistently for handwritten and generated tools where relevant.
- Make sure both stdio and streaming server code paths behave consistently.

The following is the prior conversation context from the user's chat exploration (may be truncated):

User: readonly annotaitons on readonly tools are not applied, is that right? what needs to change

Assistant: Yes — **as currently implemented, `readonly` annotations are not what drive filtering**. The filter logic is using **`http_methods`** instead.

The key behavior is here:

```python name=src/tools/tool_filter.py url=https://github.com/opensearch-project/opensearch-mcp-server-py/blob/d8f8b7429f07c45674324642fef55c4300e9ad6a/src/tools/tool_filter.py#L102-L108
def apply_write_filter(registry):
    """Apply allow_write filters to the registry."""
    for tool_name in list(registry.keys()):
        http_methods = registry[tool_name].get('http_methods', [])
        if 'GET' not in http_methods:
            registry.pop(tool_name, None)
```

So when `allow_write=False`, the code keeps only tools whose metadata includes `GET`. It does **not** inspect a `readonly` flag/annotation anywhere. And I couldn’t find any code references to `readonly` in the repo search.

## What that means

If a tool is “readonly” by annotation but its `http_methods` metadata:
- is missing,
- is not normalized,
- or includes a non-GET method pattern,

then it may still be filtered incorrectly.

Also, the current rule is a bit simplistic:
- it treats **presence of `GET`** as safe,
- rather than treating **readonly-ness** as a first-class property.

## What needs to change

You likely want to make **readonly an explicit tool metadata field** and then make `apply_write_filter()` use that field.

### Suggested change

In registry entries, add something like:

```python name=src/tools/tools.py
'ListIndexTool': {
    ...
    'readonly': True,
}
```

and for write tools:

```python name=src/tools/tools.py
'CreateQuerySetTool': {
    ...
    'readonly': False,
}
```

Then update filtering to prefer `readonly`:

```python name=src/tools/tool_filter.py
def apply_write_filter(registry):
    """Apply allow_write filters to the registry."""
    for tool_name in list(registry.keys()):
        readonly = registry[tool_name].get('readonly')
        if readonly is False:
            registry.pop(tool_name, None)
```

## Better version for backward compatibility

Since the repo already uses `http_methods`, the safest migration is:

```python name=src/tools/tool_filter.py
def apply_write_filter(registry):
    """Apply allow_write filters to the registry."""
    for tool_name in list(registry.keys()):
        tool = registry[tool_name]

        if 'readonly' in tool:
            if not tool['readonly']:
                registry.pop(tool_name, None)
            continue

        http_methods = tool.get('http_methods', [])
        if isinstance(http_methods, str):
            methods = {http_methods.upper()}
        else:
            methods = {m.upper() for m in http_methods}

        if methods != {'GET'}:
            registry.pop(tool_name, None)
```

That would:
1. honor explicit `readonly` metadata first,
2. fall back to `http_methods` for older tools.

## Why this is probably the right fix

Right now tool metadata already looks like this:

```python name=src/tools/tools.py url=https://github.com/opensearch-project/opensearch-mcp-server-py/blob/d8f8b7429f07c45674324642fef55c4300e9ad6a/src/tools/tools.py#L858-L870
'ListIndexTool': {
    'display_name': 'ListIndexTool',
    ...
    'http_methods': 'GET',
},
```

So the system’s concept of “readon...

</details>

*This pull request was created from Copilot chat.*
>


src/tools/utils.py
```diff
@@ -51,6 +51,26 @@ def load_yaml_config(filter_path):
...
+
+    Returns:
+        bool: True if the tool only uses GET, False otherwise.
+    """
+    http_methods = tool_info.get('http_methods', '')
+    if isinstance(http_methods, str):
+        methods = {m.strip().upper() for m in http_methods.split(',') if m.strip()}
+    else:
+        methods = {m.upper() for m in http_methods}
+    return methods == {'GET'}
```
arjunkumargiri: SearchIndex tool supports both GET and POST: https://github.com/opensearch-project/opensearch-mcp-server-py/blob/d8f8b7429f07c45674324642fef55c4300e9ad6a/src/tools/tools.py#L915 . 

So search index tool will not be annotated as read only.

rithinpullela: @bbartels can we mark the Search tool GET only in tools.py? 
Would appreciate it if you can check the http_method annotations for each tool and check if any need to be changed

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### Added
 
+- Adds readOnlyHint annotation to readonly tools
```
rithinpullela: Can we add PR link as well?

bbartels:
@rithinpullela @nathaliellenaa I've fixed up merge conflicts and did another pass over the pr. Thoughts? :)

src/tools/tool_filter.py
```diff
@@ -115,8 +116,7 @@ def apply_write_filter(registry, exempt_tools=None):
     for tool_name in list(registry.keys()):
         if registry[tool_name].get('bypass_write_filter') or tool_name in exempt_tools:
             continue
-        http_methods = registry[tool_name].get('http_methods', [])
-        if 'GET' not in http_methods:
+        if not is_read_only_tool(registry[tool_name]):
```
nathaliellenaa, line 119: **Issue:** This change breaks `GenericOpenSearchApiTool` functionality when `allow_write=False`.

**Before this PR:**
  - Logic: Keep tool if 'GET' in http_methods
  - `GenericOpenSearchApiTool` has http_methods='GET, POST, PUT, DELETE, HEAD, PATCH'
  - Result: ✅ Tool is kept (because it includes GET)
  
**After this PR:**
  - Logic: Keep tool only if `is_read_only_tool()` returns True
  - `GenericOpenSearchApiTool` has `read_only_hint: False` (tools.py line 1325)
  - Result: ❌ Tool is filtered out
  
**Impact:** All 6 integration tests fail because they expect `GenericOpenSearchApiTool` to be available for read operations even when `allow_write=False`.

**Why it matters:** `GenericOpenSearchApiTool` is a runtime-dispatch tool that handles both read AND write operations. Write protection happens at runtime based on the HTTP method in the request. Filtering it out entirely removes the ability to use it for GET/read operations.

src/tools/tools.py
```diff
@@ -1295,3 +1295,44 @@ async def search_experiments_tool(args: SearchExperimentsArgs) -> list[dict]:
...
+    'SampleQuerySetTool': False,
+    'DeleteQuerySetTool': False,
+    'GetExperimentTool': True,
+    'CreateExperimentTool': False,
+    'DeleteExperimentTool': False,
+    'SearchQuerySetsTool': True,
+    'SearchSearchConfigurationsTool': True,
+    'SearchJudgmentsTool': True,
+    'SearchExperimentsTool': True,
+    'GenericOpenSearchApiTool': False,
```
nathaliellenaa, line 1326: **Issue:** Marking this as False causes it to be filtered out when allow_write=False, breaking its read functionality.

**Recommendation:** Either:
  1. Mark it as bypass_write_filter: True (like memory tools do)
  2. Special-case it in the filter logic to allow it through 
  3. Consider setting read_only_hint: None and handle this case specially
  
The tool performs runtime dispatch - it's neither purely read-only nor purely write-only, so it needs special handling.

nathaliellenaa (commented):
Thanks for the PR @bbartels. CIs are failing due to some breaking changes, put the details in the comment below:

bbartels:
@nathaliellenaa all tests are passing now, thanks for the review!

bbartels:
@nathaliellenaa Gentle bump :)

bbartels:
@rithinpullela Would you mind taking a look?

bbartels:
@rithinpullela @nathaliellenaa Would you mind taking a look? :)

PR #224, opened 2026-04-16
feat(memory): Add agent memory tools with recency-aware search

arjunkumargiri:
Add SaveMemoryTool, SearchMemoryTool, and DeleteMemoryTool that give AI agents persistent cross-session memory backed by OpenSearch. Memories are stored with automatic semantic enrichment for natural language retrieval.

Key features:
- Save/search/delete memories with user, agent, and session scoping
- Recency-weighted ranking using painless script_score (exponential decay: 1-day offset, 7-day half-life)
- Auto-creation of memory index and data access policies on Serverless
- Bypass write filter so memory tools work even with allow_write=false
- Memory tools opt-in via MEMORY_TOOLS_ENABLED=true env var

Add MEMORY.md documentation covering architecture, setup for Kiro/Claude Code/Cursor, shared memory across agents, and tool reference.


src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,234 @@
...
+    },
+}
+
+# Claude Code hook definitions
+CLAUDE_CODE_SEARCH_HOOK = {
+    'matcher': '',
+    'event': 'UserPromptSubmit',
+    'type': 'command',
+    'command': (
+        'echo \'{"additionalContext": "' + SEARCH_PROMPT.replace("'", "\\'") + '"}\''
```
rithinpullela: Looks like this is not escaped properly. Could cause issues when claude code runs this.

Issue found by claude code's review:
```
Inside single-quoted bash strings, \' doesn't escape — it terminates the string. SAVE_PROMPT contains don't, so the generated command fails with /bin/sh: unexpected EOF while looking for matching ". Use shlex.quote() on the JSON payload, or write the hook as a Python one-liner: python3 -c
  'import json; print(json.dumps({"additionalContext": "..."}))'.

 Bash-only reproduction (no Python needed — paste into a shell):
  bash -c "echo '{\"additionalContext\": \"user\\'s message\"}'"
  # bash: unexpected EOF while looking for matching `"'
``

File: src/mcp_server_opensearch/installer.py [hunk 2]
```diff
@@ -0,0 +1,341 @@
...
+
+IDE_CONFIGS = {
+    'kiro': {
+        'name': 'Kiro',
+        'mcp_path': Path.home() / '.kiro' / 'settings' / 'mcp.json',
+        'steering_path': Path.home() / '.kiro' / 'steering' / 'opensearch-memory.md',
+    },
+    'claude-code': {
+        'name': 'Claude Code',
+        'mcp_path': Path.home() / '.claude' / 'settings.json',
```
rithinpullela: I believe for claude code it is`claude.json` not `settings.json`

src/mcp_server_opensearch/installer.py
```diff
@@ -0,0 +1,341 @@
...
+        aws_profile: AWS profile name, or empty for default credentials.
+        from_git: If set, use ``uvx --from "git+<url>"`` instead of the
+            PyPI release.  Useful for testing pre-release branches.
+    """
+    env = {
+        'OPENSEARCH_URL': opensearch_url,
+        'AWS_REGION': aws_region,
+        'AWS_OPENSEARCH_SERVERLESS': 'true',
+        'MEMORY_TOOLS_ENABLED': 'true',
+        'OPENSEARCH_DISABLED_CATEGORIES': 'core_tools,skills_tools',
```
rithinpullela: The category is `skills`. This would be a no op on skills_tools

src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,234 @@
...
+        'type': 'agentStop',
+    },
+    'then': {
+        'type': 'askAgent',
+        'prompt': SAVE_PROMPT,
+    },
+}
+
+# Claude Code hook definitions
+CLAUDE_CODE_SEARCH_HOOK = {
```
rithinpullela: Claude code hook Schema shape seems wrong. I experimented and check this as well. 

Here is what claude has to say for review (I validated this):
```
A Claude Code hook has two moving parts:

  1. Registration — an entry in settings.json that tells Claude Code "run this shell command when event X fires."
  2. Communication — the shell command's stdout, which Claude Code parses as JSON to decide what to do next (inject context, block the action, etc.).

  The PR gets both wrong. Either one alone would break the hook; together they break it twice.

  ---
  Layer 1 — Registration (the settings.json shape)

  What the PR writes:

  {
    "hooks": [
      {
        "matcher": "",
        "event": "UserPromptSubmit",
        "type": "command",
        "command": "echo '{\"additionalContext\": \"...\"}'"
      },
      {
        "matcher": "",
        "event": "Stop",
        "type": "command",
        "command": "echo '{\"additionalContext\": \"...\"}'"
      }
    ]
  }

  Flat list. Each entry has its own event field.

  What Claude Code actually expects:

  {
    "hooks": {
      "UserPromptSubmit": [
        {
          "matcher": "",
          "hooks": [
            {"type": "command", "command": "echo '...'"}
          ]
        }
      ],
      "Stop": [
        {
          "matcher": "",
          "hooks": [
            {"type": "command", "command": "echo '...'"}
          ]
        }
      ]
    }
  }

  Three differences:

  - hooks is a map keyed by event name, not a list. There's no event field — the event is the key.
  - Each event maps to a list of matcher groups. The matcher string filters by tool name for tool-scoped events (PreToolUse, PostToolUse); for lifecycle events like UserPromptSubmit and Stop, an empty matcher matches everything.
  - Each matcher group has its own inner hooks list of {type, command} entries. (Yes, hooks appears at two nesting levels. It's confusing but it's the actual schema.)

  Because the PR's shape doesn't match, Claude Code's config loader just doesn't recognize these entries as hooks at all — they're silently ignored at startup. No error, no warning. The command never runs.

  Downstream consequence: _hook_already_installed() at install_hooks.py:151 iterates existing_hooks expecting flat dicts with .event and .command fields. Once you fix the schema to the nested shape, that idempotency check becomes a no-op (it'll always say "not installed" because the new shape
  has no top-level event key), so running the installer twice will duplicate the hooks. The fix needs to walk the nested structure.

  ---
  Layer 2 — Communication (the stdout JSON)

  Even if Layer 1 were fixed, the hook's shell command is echo '{"additionalContext": "..."}'. The intent is: "print this JSON, Claude Code reads it, injects the text into the next turn."

  What the PR emits:

  {"additionalContext": "Before responding, search memory..."}

  What Claude Code actually reads:

  {
    "hookSpecificOutput": {
      "hookEventName": "UserPromptSubmit",
      "additionalContext": "Before responding, search memory..."
    }
  }

  The additionalContext field only works when wrapped in hookSpecificOutput. Claude Code's hook stdout parser has a small set of top-level keys it recognizes (continue, decision, reason, suppressOutput, hookSpecificOutput, systemMessage). A bare additionalContext at the top level isn't in that
   set — it's silently dropped, and the hook becomes a no-op.

  This is per-event: additionalContext is only supported for UserPromptSubmit, SessionStart, PostToolUse. For Stop, it doesn't inject context at all — you'd need a different mechanism (e.g. decision: "block" with reason) or to run the save step via a different approach entirely.

  ---
  Why both layers matter

  The "save on stop" hook is arguably broken by design even if fixed at both layers — the Stop event fires after the model has finished responding, and additionalContext doesn't apply to Stop. The PR's mental model seems to be "inject a prompt telling the model to call SaveMemoryTool," but by
  the time Stop fires there's no next turn to inject into. This probably needs to be a SessionEnd hook that runs a script which calls the MCP tool directly, or a different architecture entirely.

  So the blocker really has three sub-issues nested inside it:

  1. Wrong registration shape → hooks don't load.
  2. Wrong stdout shape → even if loaded, UserPromptSubmit hook can't inject context.
  3. Wrong event for the save flow → Stop + additionalContext is semantically incoherent.

  All three need resolution for the "hooks automate memory" story to actually work.
```

src/tools/memory_tools.py
```diff
@@ -0,0 +1,729 @@
...
+        effective_size = min(args.size, 100) if args.size else 10
+
+        # Build filter clauses
+        filters = []
+        if args.user_id:
+            filters.append({'term': {'user_id': args.user_id}})
+        if args.agent_id:
+            filters.append({'term': {'agent_id': args.agent_id}})
+        if args.session_id:
+            filters.append({'term': {'session_id': args.session_id}})
```
rithinpullela: Would it help to set these in env variables so that a user can be assured to pull exactly the memories that they want to pull?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- Add agent memory tools (`SaveMemoryTool`, `SearchMemoryTool`, `DeleteMemoryTool`) for persistent cross-session memory backed by OpenSearch with automatic semantic enrichment, recency-aware ranking, and shared memory across multiple AI agents
+- Add `MEMORY.md` documentation for agent memory setup, architecture, and tool reference
+- Add `install-hooks` CLI command to install memory lifecycle hooks for Kiro and Claude Code
+- Add `memory install` interactive CLI command for one-step setup across Kiro, Claude Code, and Cursor
```
rithinpullela: Nit: Can we add all of this as a single feature?

src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,234 @@
...
+
+import json
+import logging
+from pathlib import Path
+
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Hook definitions
```
rithinpullela: Not a blocker for this PR, can take it as a future action item. Looks like Cursor also supports hooks, should we consider adding them?

rithinpullela (commented):
Thanks for this PR @arjunkumargiri!

I think this will be a great addition to the MCP server!

Added some comments majorly related to hooks.

src/tools/memory_tools.py
arjunkumargiri: In my opinion, it might be better to control permissions using opensearch security plugin instead of defining permission control in env variable. 

I will add user_id and agent_id env variable. session_id might not be required.

src/mcp_server_opensearch/installer.py
```diff
@@ -0,0 +1,345 @@
...
+
+IDE_CONFIGS = {
+    'kiro': {
+        'name': 'Kiro',
+        'mcp_path': Path.home() / '.kiro' / 'settings' / 'mcp.json',
+        'steering_path': Path.home() / '.kiro' / 'steering' / 'opensearch-memory.md',
+    },
+    'claude-code': {
+        'name': 'Claude Code',
+        'mcp_path': Path.home() / '.claude' / 'claude.json',
```
ylwu-amzn: **[BLOCKING] Wrong Claude Code MCP config path.**

Claude Code's user-level MCP config file is `~/.claude.json` (a single file at the home root), not `~/.claude/claude.json`. Verified on a real Claude Code install on my machine: `~/.claude.json` exists and contains `mcpServers`; `~/.claude/claude.json` does not exist.

An earlier review round flagged this ("I believe for claude code it is `claude.json` not `settings.json`") and the fix moved to `~/.claude/claude.json` — still wrong.

Two correct options:
1. Write to `~/.claude.json` directly, merging into the existing JSON without clobbering the `projects` map or other top-level keys.
2. Shell out to `claude mcp add` (the officially documented API — and what `MEMORY.md` already recommends for the manual flow). Safer because Anthropic maintains the config-schema contract.

As written, end users will go through the interactive installer, see the "success" message, and find that nothing works in Claude Code.

src/tools/tool_filter.py
```diff
@@ -172,6 +178,8 @@ def process_tool_filter(
             'ExplainTool',
             'MsearchTool',
             'GenericOpenSearchApiTool',
+            'DataDistributionTool',
```
ylwu-amzn: **[BLOCKING] This reverts PR #225 silently.**

PR #225 (already on `main`) moved `DataDistributionTool` and `LogPatternAnalysisTool` **out** of `core_tools` because they require OpenSearch >= 3.3.0 and were breaking on older clusters. Re-adding them here undoes that — every user on an older cluster will hit tool-compatibility errors again.

This is very likely a rebase artifact (branch predates #225). After rebasing onto current main, these two lines should disappear from the `core_tools` list. If there's a deliberate reason to add them here, please call it out — but I don't think that's the intent given the PR is about memory tools.

src/tools/memory_tools.py
```diff
@@ -0,0 +1,748 @@
...
+            exists = await client.indices.exists(index=index_name)
+            if exists:
+                logger.debug(f'Memory index "{index_name}" already exists')
+                return
+        except Exception:
+            # If we can't check, try to create anyway
+            pass
+
+    # Index doesn't exist — create it via boto3
+    opensearch_url = os.getenv('OPENSEARCH_URL', '').strip()
```
ylwu-amzn: **[BLOCKING] Per-call connection overrides are ignored here.**

PR #231 added per-call overrides for `opensearch_url`, `aws_region`, `aws_profile`, `aws_opensearch_serverless`, etc. via `baseToolArgs`. Memory tools correctly accept `baseToolArgs` and use `get_opensearch_client(args)` elsewhere (honors overrides). But the boto3 code path in this function reads `os.getenv('OPENSEARCH_URL')` directly.

Net effect: when a user targets a different cluster via per-call params, the first `SaveMemoryTool` call will try to create the index against the wrong cluster — or fail with "Cannot determine AOSS collection ID" if `OPENSEARCH_URL` is empty on the server (which is explicitly the supported "zero-config mode" from #231).

Also, `_is_serverless(opensearch_url)` a few lines down (line 327) only checks for `aoss.` in the URL, ignoring `AWS_OPENSEARCH_SERVERLESS=true` and the new per-call `aws_opensearch_serverless` boolean. Users with VPC endpoints or custom CNAMEs will hit the wrong code path.

Fix: resolve `opensearch_url` and `is_serverless` with the same precedence as `_initialize_client_single_mode` in `src/opensearch/client.py:258–287` (args → env → headers). Ideally share a helper so this can't drift.

src/tools/tools.py
```diff
@@ -947,6 +901,7 @@ async def search_experiments_tool(args: SearchExperimentsArgs) -> list[dict]:
 # Registry of available OpenSearch tools with their metadata
 TOOL_REGISTRY = {
     **SKILLS_TOOLS_REGISTRY,
+    **MEMORY_TOOLS_REGISTRY,
```
ylwu-amzn, line 952: **[BLOCKING] Memory tools don't enforce single-mode but require single-mode env vars.**

The tools are registered in every mode (no `multi_only` or `single_only` flag), but they rely on `OPENSEARCH_URL` + AWS env vars at the server process level. In multi mode, `initialize_client` requires `args.opensearch_cluster_name` (see `src/opensearch/client.py:88`), so calling any memory tool in multi mode will fail with `"In multi mode, opensearch_cluster_name must be provided"`. The installer hardcodes a single-mode-only config, but the tools themselves stay registered when someone runs `--mode multi`.

Options:
- Add `single_only: True` to memory tool registrations (requires adding the flag mirroring `multi_only`), or
- Skip memory tool registration when `get_mode() == 'multi'`, or
- Document the restriction prominently and raise a clear error in multi mode before attempting to build the client.

This also compounds with the previous comment on `_ensure_memory_index`: in multi mode, `os.getenv('OPENSEARCH_URL')` is empty by design, so the boto3 fallback raises "Cannot determine AOSS collection ID" — not useful.

src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,469 @@
...
+def _make_python_echo_command(text: str) -> str:
+    """Build a shell command that safely prints text via python3.
+
+    Uses base64 encoding to avoid all bash quoting issues with apostrophes,
+    double quotes, or other special characters in the text.
+    """
+    import base64
+    encoded = base64.b64encode(text.encode()).decode()
+    return (
+        f'python3 -c "import base64,sys; sys.stdout.write(base64.b64decode(\'{encoded}\').decode())"'
```
ylwu-amzn: **[HIGH] Python subprocess on every hook invocation is heavyweight.**

This command fires `python3 -c "import base64,sys; sys.stdout.write(...)"` on every user prompt (UserPromptSubmit) and every Stop. That's a full Python interpreter start-up per turn — roughly 30–80 ms of added latency for Claude Code and Cursor users.

Options (in order of preference):
1. Sanitize `SEARCH_PROMPT`/`SAVE_PROMPT` to ASCII without single quotes, then `echo '{"additionalContext": "..."}'` is safe — no Python needed. The prompts are controlled by this repo, so this is not a hard constraint.
2. Ship a tiny pre-written script (`memory-hook-search.sh`) at hook-install time and reference its path in the hook command.
3. Keep Python but use `python3 -c 'print("...")'` with a statically-defined string that you know is quote-safe (still subprocess-heavy, just cleaner than base64).

src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,469 @@
...
+    import base64
+    payload = json.dumps({'decision': 'block', 'reason': reason})
+    encoded = base64.b64encode(payload.encode()).decode()
+    return (
+        f'python3 -c "import base64,sys; sys.stdout.write(base64.b64decode(\'{encoded}\').decode())"'
+    )
+
+
+CLAUDE_CODE_SEARCH_COMMAND = _make_python_echo_command(SEARCH_PROMPT)
+CLAUDE_CODE_SAVE_COMMAND = _make_claude_stop_command(SAVE_PROMPT)
```
ylwu-amzn: **[HIGH] Stop hook's `decision: "block"` has no loop guard.**

This hook emits `{"decision": "block", "reason": "...save memories using SaveMemoryTool..."}` on every Stop. Semantically, that tells Claude Code "don't stop; continue with this as the next instruction." If Claude obliges, it calls `SaveMemoryTool`, stops, Stop fires again, block fires again — infinite loop.

Cursor's equivalent (`followup_message`) has an explicit `loop_limit: 1` at line 215 that you correctly use. Claude Code's Stop hook has no such native guardrail.

You need one of:
- A deterministic "have we already saved this turn?" check (e.g., inspect recent transcript events and only emit `block` if no save has happened this turn)
- A sessionId-scoped marker file consulted by the hook
- Or drop the Stop hook and rely only on the UserPromptSubmit hook + proactive tool descriptions

As-is, end users could see their sessions appear to hang in a save loop. Please either make this bulletproof or ship without the Stop hook for Claude Code in this first iteration.

src/mcp_server_opensearch/installer.py
```diff
@@ -0,0 +1,345 @@
...
+
+    return {
+        'command': command,
+        'args': args,
+        'env': env,
+        'autoApprove': [
+            'SaveMemoryTool',
+            'SearchMemoryTool',
+            'DeleteMemoryTool',
+        ],
```
ylwu-amzn: **[HIGH] `autoApprove` is a Kiro-specific field but it's applied to all three IDE configs.**

Claude Code and Cursor do not honor `autoApprove`. Including it in the shared config means Cursor and Claude Code users get an extra unknown key (mostly harmless but confusing — users will expect the tools to be auto-approved and they won't be).

Recommend moving the `autoApprove` list into the Kiro-specific code path only.

Also worth flagging separately from a mechanical standpoint: auto-approving `SaveMemoryTool` + `DeleteMemoryTool` by default turns shared memory into a prompt-injection amplifier. If an attacker can get one agent to save attacker-controlled text, that text is retrievable by every other agent in the shared store without human approval. Suggest auto-approving `SearchMemoryTool` only (read-only), and making Save/Delete auto-approval an explicit opt-in during the interactive installer. Worth a dedicated section in MEMORY.md on "security model — memories are untrusted content."

src/tools/memory_tools.py
```diff
@@ -0,0 +1,748 @@
...
+                    'Resource': [f'collection/{collection_name}'],
+                    'Permission': ['aoss:*'],
+                    'ResourceType': 'collection',
+                },
+                {
+                    'Resource': [f'index/{collection_name}/*'],
+                    'Permission': ['aoss:*'],
+                    'ResourceType': 'index',
+                },
+            ],
```
ylwu-amzn, line 286: **[HIGH] AOSS data access policy is broader than needed.**

The auto-created policy grants `aoss:*` over the entire collection (`collection/{name}` and `index/{name}/*`), not just the memory index. If the user's collection hosts other indices, this policy now grants the caller full admin over all of them.

Suggest:
- Scope the `index` rule to `index/{collection_name}/{MEMORY_INDEX_NAME}` (or a `mcp-memory-*` prefix), not `index/{collection_name}/*`.
- Narrow the `Permission` list from `aoss:*` to the specific permissions needed (create, read, write on that one index).
- At minimum, log a prominent warning when creating the policy so operators notice the scope.

This is a one-shot silent side-effect on the operator's AWS account. Even if the operator has strong IAM boundaries, the policy itself becomes a discoverability surprise.

src/mcp_server_opensearch/install_hooks.py
```diff
@@ -0,0 +1,469 @@
...
+                    try:
+                        # Extract the base64 token from the command string
+                        import re
+                        match = re.search(r"b64decode\('([A-Za-z0-9+/=]+)'\)", command)
+                        if match:
+                            decoded = base64.b64decode(match.group(1)).decode()
+                            if 'SearchMemoryTool' in decoded or 'SaveMemoryTool' in decoded:
+                                return True
+                    except Exception:
+                        pass
```
ylwu-amzn: **[MEDIUM] Silent catch-all hides idempotency failures.**

```python
except Exception:
    pass
```

If the regex or base64 decode fails, we silently treat the hook as "not installed" and end up duplicating it on the next run. At minimum, log at debug level (`logger.debug`) so an operator debugging a duplicated-hooks bug can see why the already-installed check missed.

Same pattern at the Cursor check below (`_cursor_hooks_already_installed` ~line 375).

src/tools/memory_tools.py
```diff
@@ -0,0 +1,748 @@
...
+        # prioritized while highly relevant older memories still surface.
+        if not is_list_all:
+            recency_script = (
+                'double relevance = _score;'
+                'long now = new Date().getTime();'
+                'long created = doc[\'created_at\'].value.toInstant().toEpochMilli();'
+                'double ageHours = (now - created) / 3600000.0;'
+                'double decay = Math.exp(-0.693 * Math.max(0, ageHours - 24) / 168.0);'
+                'return relevance * decay;'
+            )
```
ylwu-amzn, line 659: **[MEDIUM] Painless `new Date().getTime()` makes the query uncacheable.**

Pulling "now" from inside the script means every `SearchMemoryTool` request produces a unique script output, defeating shard-level query caching. For large memory stores this becomes measurable.

Alternatives:
- Pass `now_ms` as a script parameter bound by the Python layer at query build time. This still defeats caching *across* calls but is cleaner to reason about.
- Use `function_score` with `gauss(created_at, origin=now, scale=7d)` — OpenSearch handles the decay natively and the query structure is more idiomatic than a custom painless script.

Not a correctness bug — worth a comment acknowledging the trade-off or a refactor to `function_score`.

src/tools/memory_tools.py
```diff
@@ -0,0 +1,748 @@
...
+            ),
+            'input_schema': DeleteMemoryArgs.model_json_schema(),
+            'function': delete_memory_tool,
+            'args_model': DeleteMemoryArgs,
+            'min_version': '1.0.0',
+            'http_methods': 'GET, DELETE',
+            'bypass_write_filter': True,
+        },
+    }
+    if _is_memory_enabled()
```
ylwu-amzn, line 820: **[LOW] Module-level side effect: `MEMORY_TOOLS_REGISTRY` is frozen at import time.**

```python
MEMORY_TOOLS_REGISTRY = (
    {...}
    if _is_memory_enabled()
    else {}
)
```

The tests work around this by deleting `tools.memory_tools` from `sys.modules` between tests (see `tests/tools/test_memory_tools.py:17–21`). That's a symptom — if someone later tests "flip the env var at runtime" they'll be confused.

Consider exposing a `get_memory_tools_registry()` function and calling it from `tools.py` when composing `TOOL_REGISTRY`. Tests get simpler and the import-order coupling goes away.

tests/tools/test_memory_tools.py
```diff
@@ -0,0 +1,734 @@
...
+            user_id='alice',
+            tags='preference,ui',
+            opensearch_cluster_name='',
+        )
+
+        with (
+            patch('tools.memory_tools._ensure_memory_index', new_callable=AsyncMock),
+            patch('opensearch.client.get_opensearch_client', return_value=mock_client),
+        ):
+            result = await self._save_memory_tool(args)
```
ylwu-amzn, line 204: **[LOW] Test patches `get_opensearch_client` with `return_value=mock_client`, not an async context manager.**

The code under test does `async with get_opensearch_client(args) as client`. The patch here replaces `get_opensearch_client` such that calling it returns `mock_client` directly, not a context manager. This only works because `mock_client` is an `AsyncMock` with `__aenter__`/`__aexit__` set up on the same mock — so the `async with` happens to find those methods on the mock itself.

Fragile — if anyone refactors the mock to be a plain `Mock` with a `.client = AsyncMock()` attribute, these tests break in a confusing way.

Suggest a small helper:

```python
def _cm(mock_client):
    cm = AsyncMock()
    cm.__aenter__.return_value = mock_client
    cm.__aexit__.return_value = False
    return cm

with patch('opensearch.client.get_opensearch_client', return_value=_cm(mock_client)):
    ...
```

MEMORY.md
```diff
@@ -0,0 +1,299 @@
...
+AI agents lose context between sessions. Every new conversation starts from scratch — the agent doesn't remember your project conventions, past decisions, or debugging history. Agent Memory solves this by giving agents a persistent knowledge store they can read from and write to during conversations.
+
+**Key benefits:**
+
+- **Cross-session continuity** — agents remember decisions, preferences, and project context across conversations
+- **Shared memory across agents** — because memory lives in OpenSearch, multiple agents (Kiro, Claude Code, Cursor, or your own custom agents) can read and write to the same memory store
+- **Semantic search** — agents find relevant memories using natural language, not exact keyword matches, powered by OpenSearch automatic semantic enrichment
+- **Recency-aware ranking** — search results blend semantic relevance with recency so recent memories are prioritized while highly relevant older memories still surface
+- **Scoped access** — memories can be scoped by user, agent, or session to control visibility
+- **No LLM dependency** — memory storage and retrieval use OpenSearch directly with no additional LLM calls for embeddings
```
ylwu-amzn: **[LOW / docs] "No LLM dependency" is misleading.**

AWS "automatic semantic enrichment" is backed by ML connectors and an embedding / sparse-encoder model managed by OpenSearch Service. There's definitely a model in the loop — it's just that it's run on the OpenSearch side rather than the MCP-server side. The phrase "No LLM dependency" implies zero embedding cost, which won't match operator expectations.

Suggest rewording to "No MCP-side LLM calls" or "No embedding setup required" — both true, neither overstating.

src/mcp_server_opensearch/__init__.py
```diff
@@ -37,11 +37,65 @@ def parse_unknown_args_to_dict(unknown_args: List[str]) -> Dict[str, str]:
...
+    # Check for subcommands before full arg parsing
+    import sys
+
+    if len(sys.argv) > 2 and sys.argv[1] == 'memory' and sys.argv[2] == 'install':
+        _run_memory_install()
+        return
+
+    if len(sys.argv) > 1 and sys.argv[1] == 'install-hooks':
+        _run_install_hooks(sys.argv[2:])
+        return
```
ylwu-amzn, line 106: **[LOW] `install-hooks` and `memory install` subcommands aren't discoverable via `--help`.**

They're parsed via special-cased `sys.argv[1] == 'install-hooks'` / `sys.argv[1:3] == ['memory', 'install']` branches before the main argparse runs. Users running `opensearch-mcp-server-py --help` won't see either. A small argparse `subparsers` migration would make both visible without changing behavior.

src/tools/memory_tools.py
nathaliellenaa: +1 This might lead to security risk due to overly permissive permissions. Suggest to replace it with specific, minimal permissions if possible

src/tools/memory_tools.py
```diff
@@ -0,0 +1,748 @@
...
+        # recency. Uses exponential decay: memories within 1 day get full score,
+        # then decay to 50% at 7 days. This ensures recent memories are
+        # prioritized while highly relevant older memories still surface.
+        if not is_list_all:
+            recency_script = (
+                'double relevance = _score;'
+                'long now = new Date().getTime();'
+                'long created = doc[\'created_at\'].value.toInstant().toEpochMilli();'
+                'double ageHours = (now - created) / 3600000.0;'
+                'double decay = Math.exp(-0.693 * Math.max(0, ageHours - 24) / 168.0);'
```
nathaliellenaa: This value is hardcoded, which may not be suitable for all use cases. Should we make it configurable?

src/tools/memory_tools.py
arjunkumargiri: Good point, the changes will be rebased after PR 231 is merged.

ylwu-amzn:
LGTM. We can approve once conflicts resolved

ylwu-amzn (approved):

rithinpullela (approved):

PR #231, opened 2026-04-22
feat: Add dynamic per-call connection parameters for multi-tenant support

### LINKED ISSUE #230: [FEATURE] Multi-tenant support: allow agents to pass connection parameters per tool call
## Is your feature request related to a problem?

Today the OpenSearch MCP server operates as a single-tenant setup — users must hardcode an `OPENSEARCH_URL` and authentication credentials at server startup via environment variables or a YAML config file. This means:

- **One server instance = one cluster.** To work with multiple clusters, users must either run multiple server instances or pre-configure every cluster in a multi-mode YAML file.
- **Agents cannot dynamically choose which cluster to talk to.** The connection target is fixed at setup time, not at query time.
- **Onboarding friction.** Users must know their cluster endpoints upfront and configure them before the server can be useful.

## Proposed solution

Make the MCP server multi-tenant by allowing agents to pass connection parameters (endpoint URL, auth credentials, region, etc.) as optional fields on every tool call. The server would then establish or reuse connections dynamically based on the parameters provided.

The `baseToolArgs` class already defines optional override fields (`opensearch_url`, `aws_profile`, `aws_region`, etc.) on every tool. The change is to fully support these as first-class connection parameters so that:

1. **No pre-configuration is required.** An agent can call `ListIndexTool` with `opensearch_url`, `aws_profile`, and `aws_region` and it just works — no env vars or YAML needed.
2. **Agents can target any cluster in a single conversation.** For example, list indices on cluster A, then search on cluster B, all through the same server instance.
3. **Connection pooling by endpoint.** The server should cache and reuse clients keyed by the effective connection parameters to avoid reconnecting on every call.

### Flow 1: User tells the agent which cluster to query

```
User: "List all indices on my production cluster at https://prod.us-east-1.es.amazonaws.com"

Agent calls ListIndexTool with:
  opensearch_url: "https://prod.opensearch.com"
  aws_profile: "prod-readonly"
  aws_region: "us-east-1"
```

### Flow 2: Agent discovers endpoints from an SOP or knowledge base

```
User: "Check the health of all our OpenSearch clusters"

Agent looks up the team SOP / knowledge base / runbook and finds:
  - prod cluster: https://prod.opensearch.com (us-east-1)
  - staging cluster: https://staging.opensearch.com (us-west-2)

Agent calls ClusterHealthTool twice, once per cluster:
  Call 1: opensearch_url="https://prod.opensearch.com", aws_region="us-east-1"
  Call 2: opensearch_url="https://staging.opensearch.com", aws_region="us-west-2"

Agent synthesizes results: "Prod is green (5 nodes), staging is yellow (2 nodes, 3 unassigned shards)."
```

### Flow 3: Agent discovers all collections in an AWS account

```
User: "List indices across all my OpenSearch Serverless collections"

Agent uses AWS CLI / SDK to call ListCollections, gets endpoints for each collection,
then calls ListIndexTool for each:
  Call 1: opensearch_url="https://abc123.us-east-1.aoss.amazonaws.com", aws_opensearch_serverless=true
  Call 2: opensearch_url="https://def456.us-east-1.aoss.amazonaws.com", aws_opensearch_serverless=true

Agent returns a consolidated view of all indices across all collections.
```

## Why this matters

- **Agentic workflows are dynamic.** Agents operating in multi-cluster environments (e.g., prod + staging, or multiple Serverless collections) need to route requests at runtime, not at config time.
- **Reduces setup burden.** Users can start using the MCP server without knowing every endpoint upfront. The agent figures it out.
- **Enables richer automation.** Cross-cluster comparisons, health checks across environments, and migration workflows all become possible through a single server instance.

## Additional context

The optional override fields on `baseToolArgs` already exist in the codebase. This issue is about ensuring the full connection lifecycle (client creation, caching, auth resolution) works end-to-end when these fields are provided without any pre-configured environment variables or YAML config.

arjunkumargiri:
Enable agents to pass connection parameters (opensearch_url, aws_profile, aws_region, etc.) directly in each tool call, allowing a single MCP server instance to dynamically target different OpenSearch clusters without requiring pre-configured environment variables or server restarts.

Token optimization: connection override fields are conditionally exposed in tool schemas based on whether a connection is pre-configured:
- When OPENSEARCH_URL is set (single mode) or clusters are loaded from YAML (multi mode): override fields are hidden from schemas (zero token overhead), since the agent doesn't need to provide them.
- When nothing is configured (zero-config mode): override fields are exposed in schemas and server-level MCP instructions guide the LLM on their usage.

Changes:
- Add optional connection override fields to baseToolArgs (tool_params.py)
- Apply per-call overrides in single mode client initialization (client.py)
- Conditionally expose override fields in tool schemas (tool_filter.py)
- Add server-level instructions for zero-config mode (server_instructions.py)
- Pass instructions to MCP Server in both stdio and streaming servers
- Add comprehensive tests for connection overrides, instructions, and conditional schema stripping across single/multi modes
- Update README.md and USER_GUIDE.md with documentation

Resolves #230


src/mcp_server_opensearch/stdio_server.py
```diff
@@ -32,7 +33,7 @@ async def serve(
     if config_file_path:
         set_config_file_path(config_file_path)
 
-    server = Server('opensearch-mcp-server')
+    server = Server('opensearch-mcp-server', instructions=get_server_instructions())
```
rithinpullela: This check should be after like 38 since the cluster_registry variable is not loaded yet. So although the yml file has cluster config information, it assumes nothing is setup and starts the MCP server in dynamic mode.

rithinpullela: This has been handled well in the streaming server.

USER_GUIDE.md
```diff
@@ -211,6 +212,94 @@ The LLM needs to have context about the available cluster names to make informed
 
 The LLM should choose the appropriate cluster based on the operation context (e.g., use `local-dev` for testing, `production` for production data).
 
+## Dynamic Connection Parameters
+
+All tools accept optional connection parameters that override the server's configured environment variables (single mode) or cluster config (multi mode) on a per-call basis. This enables agents to dynamically target different OpenSearch clusters without restarting or reconfiguring the server.
```
rithinpullela: Looks like we never enter the dynamic connection mode if we see Opensearch URL env variable or if we see multi mode. Should we reword this to say this applies to everything but Openserach url env variable? 

I believe we should also call out that this will kick in automatically when there is no URL or yml configured.

Also nit: would it make sense to expose an env variable which can control dynamic mode?

rithinpullela (dismissed):
Few small concerns called out.

Can we add some Integration tests for this new mode? 
We can add ITs for:
 1. Zero-config single mode, no env vars, and validate MCP server exposing all variables in tools + provided variables taking precedence over corresponding env variables (except OpenSearch URL variable which is required)
2. -ve test cases: When URL is set, the additional variables are not exposed

integration_tests/tools/test_dynamic_connection.py
```diff
@@ -0,0 +1,164 @@
...
+    async def test_inline_params_take_precedence_over_env_vars(self, zero_config_server):
+        """Per-call opensearch_url must override any env var that might be set.
+
+        The zero_config_server has no env vars set, so the only way this
+        succeeds is if the per-call params are actually used.
+        """
+        call_args = _build_inline_call_args()
+        async with mcp_client(zero_config_server.url) as session:
+            result = await session.call_tool('ClusterHealthTool', arguments=call_args)
+            assert_tool_success(result)
```
rithinpullela: The test name here is misleading, this test is doing the same thing as the above test and not actually testing if the inline variables take precedence or not

rithinpullela (dismissed):
Added a small nit on IT, overall LGTM!

src/opensearch/client.py
```diff
@@ -81,8 +81,8 @@ def initialize_client(args: baseToolArgs) -> AsyncOpenSearch:
         logger.info(f'Initializing OpenSearch client in {mode} mode')
 
         if mode == 'single':
-            # In single mode, always use environment variables, ignore cluster name
-            return _initialize_client_single_mode()
+            # In single mode, use environment variables with optional per-call overrides from args
+            return _initialize_client_single_mode(args)
```
ylwu-amzn, line 85: **Schema / runtime divergence in multi mode.** `_initialize_client_single_mode` now reads `args.opensearch_url`, `args.aws_region`, etc. and applies them as per-call overrides — great. But `_initialize_client_multi_mode(cluster_info)` at line 358 does **not** accept `args` at all, so the same override values are silently ignored in multi mode.

Combined with `tool_filter.py`'s multi-mode branch that still strips/exposes `CONNECTION_OVERRIDE_FIELDS` based on `is_dynamic_mode_enabled()`, this means:

1. If someone runs multi mode with `OPENSEARCH_DYNAMIC_CONNECTION=true`, the override fields appear in every tool schema, but the values the agent passes are dropped on the floor at client init time.
2. `python -m mcp_server_opensearch --mode multi` with no config file auto-enters dynamic mode (no clusters loaded, no `OPENSEARCH_URL`) — and then `initialize_client` at line 88 hard-requires `opensearch_cluster_name` anyway.

Two options:
- **Preferred:** thread `args` into `_initialize_client_multi_mode` and apply the same overrides there, so schema and runtime agree.
- **Alternative:** explicitly scope this feature to single mode — skip the multi-mode branch in `tool_filter.get_tools` and update the 'Notes' in USER_GUIDE.md (which already says "available in single mode").

The current state is confusing: the schema advertises fields that may silently no-op.

src/opensearch/client.py
```diff
@@ -245,13 +254,37 @@ def _initialize_client_single_mode() -> AsyncOpenSearch:
                 logger.warning(
                     f'Invalid OPENSEARCH_MAX_RESPONSE_SIZE format: {max_response_size_str}, using default'
                 )
+
+        # Apply per-call overrides from tool args (if provided)
+        if args is not None:
+            if args.opensearch_url is not None:
+                opensearch_url = args.opensearch_url.strip()
+            if args.opensearch_username is not None:
+                opensearch_username = args.opensearch_username
```
ylwu-amzn: Nit: every other string override here calls `.strip()` (url, iam_arn, profile, aws_region). Username probably should too for consistency. Password deliberately should not — leading/trailing whitespace in passwords is legal — so that one is fine as-is, just worth a brief comment explaining why if kept intentional.

USER_GUIDE.md
```diff
@@ -211,6 +212,117 @@ The LLM needs to have context about the available cluster names to make informed
...
+`OPENSEARCH_DYNAMIC_CONNECTION` takes precedence over auto-detection. Use it to explicitly enable dynamic mode when you want agents to override a pre-configured URL, or to explicitly disable it in zero-config deployments where you don't want the override fields exposed.
+
+### How It Works
+
+When a tool is called, any connection parameter provided in the tool input takes precedence over the corresponding environment variable. Parameters that are omitted (or set to `null`) fall back to the server's existing configuration.
+
+**Priority order (highest to lowest):**
+1. Tool input parameters (per-call overrides)
+2. Header-based authentication values (when `OPENSEARCH_HEADER_AUTH=true`)
+3. Environment variables / cluster YAML config
```
ylwu-amzn: The `### How It Works` section is duplicated — the same heading and the same "Priority order" list appear at lines 221–228 and again here at 242–249. Looks like a rebase artifact from when the `OPENSEARCH_DYNAMIC_CONNECTION` env-var docs were inserted. Please delete one of the two copies.

src/mcp_server_opensearch/stdio_server.py
```diff
@@ -32,11 +33,13 @@ async def serve(
     if config_file_path:
         set_config_file_path(config_file_path)
 
-    server = Server('opensearch-mcp-server')
-    # Load clusters from YAML file
+    # Load clusters from YAML file before evaluating server instructions,
+    # so has_preconfigured_connection() sees the populated cluster registry.
     if mode == 'multi':
         await load_clusters_from_yaml(config_file_path)
```
ylwu-amzn, line 38: Good fix for the ordering issue the earlier review flagged. One edge case to sanity-check: in multi mode **without** a config file (the documented "Fallback to single mode behavior" path in USER_GUIDE.md), `load_clusters_from_yaml('')` returns early without loading anything, so `has_preconfigured_connection()` returns False and the server enters dynamic mode. Combined with the multi-mode override gap I commented on in `client.py`, this surfaces override fields that then silently no-op. Worth a quick IT run for `--mode multi` without `--config`.

src/mcp_server_opensearch/server_instructions.py
```diff
@@ -0,0 +1,105 @@
...
+    'opensearch_username',
+    'opensearch_password',
+    'opensearch_no_auth',
+    'aws_region',
+    'aws_iam_arn',
+    'aws_profile',
+    'aws_opensearch_serverless',
+    'opensearch_ssl_verify',
+    'opensearch_timeout',
+}
```
ylwu-amzn: Nit — `CONNECTION_OVERRIDE_FIELDS` is a hand-maintained set that mirrors the non-`opensearch_cluster_name` fields in `baseToolArgs`. If someone adds an 11th override to `baseToolArgs` (or this list) but forgets the other, schema stripping will silently mismatch and existing tests will still pass.

Consider deriving it once:

```python
from tools.tool_params import baseToolArgs
_NON_OVERRIDE_FIELDS = {'opensearch_cluster_name'}
CONNECTION_OVERRIDE_FIELDS = set(baseToolArgs.model_fields) - _NON_OVERRIDE_FIELDS
```

Non-blocking — just removes a whole class of drift bugs.

src/mcp_server_opensearch/server_instructions.py
```diff
@@ -0,0 +1,105 @@
...
+
+_DYNAMIC_CONNECTION_INSTRUCTIONS = """\
+This OpenSearch MCP server has no pre-configured endpoint. \
+You must provide connection parameters on each tool call.
+
+Every tool accepts these optional connection parameters:
+- opensearch_url (required): OpenSearch endpoint URL
```
ylwu-amzn, line 29: Nit: the instruction string says `opensearch_url (required)` but `baseToolArgs.opensearch_url` is `Optional[str] = None` and is not in the schema's `required` array. In practice the LLM reads this as a hint and complies, but a strict MCP client that validates against the schema would accept a call without it, then fail at runtime with the existing `OPENSEARCH_URL is required` error.

Consider either (a) adding `opensearch_url` to the `required` array when dynamic mode is active, or (b) rewording to something like "you must provide `opensearch_url` — the call will fail without it".

src/tools/tool_filter.py
```diff
@@ -398,12 +411,24 @@ async def get_tools(tool_registry: dict, config_file_path: str = '') -> dict:
...
+        # doesn't need to provide them. When nothing is configured (zero-config
+        # / multi-tenant mode), keep them so agents can pass endpoints dynamically.
         schema = tool_info['input_schema'].copy()
         if 'properties' in schema:
-            base_fields = baseToolArgs.model_fields.keys()
-            for field in base_fields:
+            from mcp_server_opensearch.server_instructions import (
+                CONNECTION_OVERRIDE_FIELDS,
+                is_dynamic_mode_enabled,
+            )
```
ylwu-amzn: Nit: `from mcp_server_opensearch.server_instructions import (...)` is imported both here (line 422–425) and in the multi-mode branch at line 358–361. The inline imports dodge a circular import, which is fine, but the duplication is avoidable — either hoist it to a single local import at the top of `get_tools`, or keep it module-level if the circular-import risk turns out to be imaginary.

tests/opensearch/test_connection_overrides.py
```diff
@@ -0,0 +1,569 @@
...
+            'opensearch_username',
+            'opensearch_password',
+            'opensearch_no_auth',
+            'aws_region',
+            'aws_iam_arn',
+            'aws_profile',
+            'aws_opensearch_serverless',
+            'opensearch_ssl_verify',
+            'opensearch_timeout',
+        }
```
ylwu-amzn: Nit: this test re-defines the override field set inline instead of importing `CONNECTION_OVERRIDE_FIELDS` from `mcp_server_opensearch.server_instructions`. If that constant drifts (new field added), this test will still pass even though coverage has regressed. The integration test at `integration_tests/tools/test_dynamic_connection.py:20` already imports the constant — suggest doing the same here.

src/opensearch/client.py
rithinpullela: +1, found the same behavior. 

I think in multi mode we should not expose these variables. The feature should be confined to single mode.
In ` tool_filter.py:357-371`, we should not gate with this: `if not is_dynamic_mode_enabled():`

arjunkumargiri: ack, will remove multi mode support, it could be added later if required.

src/mcp_server_opensearch/stdio_server.py
arjunkumargiri: Multi mode support is now decoupled from this PR. Will roll back these changes.

rithinpullela (commented):
LGTM!
Thanks for the PR @arjunkumargiri 
Can you resolve the merge conflict? I can merge after that

rithinpullela (approved):
LGTM! Thanks Arjun!

PR #253, opened 2026-05-21
Default streamable HTTP --host to 127.0.0.1

rithinpullela:
## Summary

- Change default `--host` for the `stream` transport from `0.0.0.0` to `127.0.0.1` so the streamable HTTP listener binds to loopback by default.
- Operators who need to expose the listener on other interfaces opt in explicitly with `--host 0.0.0.0`.
- Update `USER_GUIDE.md` to document the new default and call out that the listener has no built-in inbound auth — operators binding to non-loopback should front it with an authenticating proxy or restrict access via network controls.

## Why

The MCP server's streamable HTTP listener has no built-in client→server authentication. The MCP authorization spec (2025-06-18) makes inbound auth optional, and almost every comparable self-hosted MCP server (official references, `elastic/mcp-server-elasticsearch`, `awslabs/mcp`, etc.) ships none. Defaulting `--host` to `0.0.0.0` means an operator running `opensearch-mcp-server --transport stream` without reading the flags exposes every registered tool on every interface the host can reach.

This matches the posture of `mongodb-js/mongodb-mcp-server`, which binds HTTP to `127.0.0.1` by default and documents that operators must front it with a gateway for production use. The flag itself is unchanged — only the default flips.

## Test plan

- [x] `opensearch-mcp-server --transport stream` binds to `127.0.0.1:9900`
- [x] `opensearch-mcp-server --transport stream --host 0.0.0.0` still binds to all interfaces
- [x] Existing unit + integration tests pass (no test asserts the old default)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 - Add OpenSearch mTLS support for single-cluster and multi-cluster configurations, including CA bundle, client certificate, and client key settings
 
+### Changed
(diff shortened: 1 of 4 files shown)


nathaliellenaa (approved):
LGTM

PR #255, opened 2026-05-22
feat: add connection pooling for OpenSearch clients

paulo-fogaca-sinch:
Implement client caching to improve performance by reusing connections across requests, avoiding repeated SSL handshakes and connection establishment.

Changes:
- Add client_cache.py module with caching functionality
- Update client.py to use cached clients and async client creation
- Add cleanup hooks in stdio_server.py and streaming_server.py
- Clients are cached based on configuration hash and reused
- Proper cleanup on server shutdown via close_all_clients()

### Description
_Describe what this change achieves._

### Issues Resolved

### DIFF
--- a/src/opensearch/client_cache.py
+++ b/src/opensearch/client_cache.py
@@ -1,0 +1,175 @@
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""OpenSearch client caching for connection pooling and performance.
+
+This module provides a global cache for OpenSearch clients to enable connection
+reuse across multiple requests, significantly improving performance by avoiding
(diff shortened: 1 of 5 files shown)


src/opensearch/client_cache.py
```diff
@@ -0,0 +1,175 @@
...
+import logging
+import os
+from opensearchpy import AsyncOpenSearch
+from typing import Dict, Optional
+
+
+logger = logging.getLogger(__name__)
+
+# Global client cache
+_client_cache: Dict[str, AsyncOpenSearch] = {}
```
Hailong-am, line 23: do we have size limit and ttl?

PR #256, opened 2026-05-22
Add per-category write permissions

nathaliellenaa:
### Problem
 The `OPENSEARCH_SETTINGS_ALLOW_WRITE` setting is a global toggle. When set to false, it blocks write operations across all tools — including tools in categories that require write access to maximize their functionality (e.g., search relevance). To enable those tools, users must set `allow_write=true`, which also unlocks write operations for `GenericOpenSearchApiTool` (arbitrary index deletion, bulk writes, etc.), creating an all-or-nothing trade-off between safety and functionality.

### Solution

Introduce `allow_write_categories` — a new setting that selectively enables write operations for specific tool categories while keeping the global write protection active. For example:
**YAML config:**
```yaml
  tool_filters:
    settings:
      allow_write: false
      allow_write_categories:
        - search_relevance
```

  **Environment variable:**
```
  export OPENSEARCH_SETTINGS_ALLOW_WRITE=false
  export OPENSEARCH_SETTINGS_ALLOW_WRITE_CATEGORIES="search_relevance"
  ```
This keeps `GenericOpenSearchApiTool` write operations blocked at runtime, while `search_relevance` tools with read and write operations remain in the registry and fully functional.

### Note
- When both config file and environment variables are set, the config file takes precedence — consistent with existing tool filtering behavior (https://github.com/opensearch-project/opensearch-mcp-server-py/blob/main/USER_GUIDE.md#important-notes)

### Issues Resolved

### DIFF
--- a/src/tools/tool_filter.py
+++ b/src/tools/tool_filter.py
@@ -21,6 +21,9 @@ from opensearch.helper import get_opensearch_version
 # This is set during server initialization and used by individual tools
 _resolved_allow_write_setting = None
 
+# Global variable to store the resolved allow_write_categories setting
+_resolved_allow_write_categories = None
+
 
 def process_regex_patterns(regex_list, tool_names):
     """Process regex patterns and return matching tool names."""
@@ -44,6 +47,36 @@ def set_allow_write_setting(allow_write: bool) -> None:
     global _resolved_allow_write_setting
     _resolved_allow_write_setting = allow_write
     logging.debug(f'Set global allow_write setting to: {allow_write}')
+
+
+def set_allow_write_categories(categories: list) -> None:
+    """Set the global allow_write_categories setting.
+
+    When allow_write is false, tools belonging to these categories are still
+    permitted to perform write operations. This allows granular control over
+    which tool categories can write while keeping the global default restrictive.
+
+    Args:
+        categories: List of category names allowed to perform write operations
+    """
+    global _resolved_allow_write_categories
+    _resolved_allow_write_categories = categories
+    logging.debug(f'Set allow_write_categories to: {categories}')
+
+
+def get_allow_write_categories() -> list:
+    """Get the allow_write_categories setting.
+
+    Returns:
+        list: List of category names allowed to perform write operations, or empty list if not set
+    """
+    global _resolved_allow_write_categories
+    if _resolved_allow_write_categories is not None:
+        return _resolved_allow_write_categories
+    env_value = os.getenv('OPENSEARCH_SETTINGS_ALLOW_WRITE_CATEGORIES', '')
+    if env_value:
+        return parse_comma_separated(env_value)
+    return []
 
 
 def get_allow_write_setting() -> bool:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/tools/tool_filter.py [hunk 1]
```diff
@@ -100,14 +133,48 @@ def _resolve_allow_write_setting(config_file_path: str = None) -> bool:
     return allow_write
 
 
-def apply_write_filter(registry):
+def _resolve_allow_write_categories(config_file_path: str = None) -> list:
```
mingshl: the config file is going to overwrite the environment variables. The source of the truth is the config file in this case. The precedence is: config file > env var > default. I think it makes sense but worth mentioning in the PR description.

src/tools/tool_filter.py
```diff
@@ -311,7 +382,22 @@ def process_tool_filter(
 
         # Apply allow_write filter first
         if not allow_write:
-            apply_write_filter(tool_registry)
+            exempt_tools = set()
+            if allow_write_categories:
+                display_to_key = {
+                    info.get('display_name', '').lower(): key
```
mingshl, line 327: this is case sensitive from info.

src/tools/tool_filter.py
nathaliellenaa: Correct, the config file has higher priority than the env variables. This is consistent with the existing tool filtering feature. Updated PR description with this info
https://github.com/opensearch-project/opensearch-mcp-server-py/blob/main/USER_GUIDE.md#important-notes:
> When both config file and environment variables are provided, the config file will be prioritized

src/tools/tool_filter.py
nathaliellenaa: The lookup is case-insensitive. The map keys and the search terms are both lowercased before comparison. The category list and `display_to_key` map both read from `tool_registry[name].get('display_name')`, so they always stay in sync, even with custom display names. This is the same pattern used by the existing enabled/disabled tool filtering.

src/tools/tool_filter.py
```diff
@@ -100,14 +133,48 @@ def _resolve_allow_write_setting(config_file_path: str = None) -> bool:
     return allow_write
 
 
-def apply_write_filter(registry):
+def _resolve_allow_write_categories(config_file_path: str = None) -> list:
+    """Resolve the allow_write_categories setting from environment variable or config file.
+
+    Args:
```
rithinpullela: This is not used in the startup path, we resolve the tools separately in the process_tool_filter function, this is essentially dead code which is not used.

src/tools/tool_filter.py
```diff
@@ -406,6 +492,10 @@ async def get_tools(tool_registry: dict, config_file_path: str = '') -> dict:
     resolved_allow_write = _resolve_allow_write_setting(config_file_path)
     set_allow_write_setting(resolved_allow_write)
 
+    # Resolve and set the global allow_write_categories setting
+    resolved_categories = _resolve_allow_write_categories(config_file_path)
+    set_allow_write_categories(resolved_categories)
```
rithinpullela: Linking to above comment, we do set a global variable here, but it is never used for filtering

rithinpullela (commented):
Functionality looks good, but found some dead code

src/tools/tool_filter.py
nathaliellenaa: Removed it

src/tools/tool_filter.py
nathaliellenaa: Removed it

rithinpullela (approved):

PR #279, opened 2026-06-25
feat: fallback to _resolve/index when _cat/indices returns 403

Hailong-am:
## Summary
- When users lack `cluster_monitor`/`indices_monitor` permissions, `_cat/indices` returns 403
- Adds transparent fallback to `GET /_resolve/index/*` which only requires index-level read permissions (included in `read`/`search` action groups)
- When fallback is used with `include_detail=True`, the response is annotated so the LLM knows health, size, and doc count are unavailable
- When `include_detail=False` (names only), no annotation needed since the data is equivalent

## Permission comparison

| | `_cat/indices` | `_resolve/index/*` |
|---|---|---|
| Cluster perms | `cluster:monitor/health`, `cluster:monitor/state` | None |
| Index perms | `indices:monitor/stats` | `indices:admin/resolve/index` |
| Action groups | `cluster_monitor` + `indices_monitor` | `read` or `search` |
| Info returned | doc count, store size, health, status, shards | index names, aliases, data streams |

## Test plan
- [x] Existing unit tests pass (527 passed)
- [x] New unit tests for fallback path (403 triggers resolve_index)
- [x] New unit test for pattern passthrough to resolve_index
- [ ] Integration test with a restricted-permissions user


Hailong-am:
@nathaliellenaa @rithinpullela would you mind help me to review it

src/opensearch/helper.py
```diff
@@ -64,15 +64,30 @@
...
-        # Pass index parameter if provided to filter results by pattern or specific index
         index_param = args.index if args.index else None
-        response = await client.cat.indices(index=index_param, format='json')
-        return response
+        try:
+            response = await client.cat.indices(index=index_param, format='json')
+            return response, False
+        except AuthorizationException:
+            resolve_pattern = index_param if index_param else '*'
+            resolve_response = await client.indices.resolve_index(resolve_pattern)
```
nathaliellenaa: The async signature is `resolve_index(self, *, name, params=None, headers=None)` — name is keyword-only ([ref](https://github.com/opensearch-project/opensearch-py/blob/3.1/opensearchpy/_async/client/indices.py#L2665)). The positional call raises:
```
TypeError: IndicesClient.resolve_index() takes 1 positional argument but 2 positional arguments (and 2 keyword-only arguments) were given
```
So on a real 403, the fallback throws before it can list anything; the `TypeError` propagates to the except `Exception` in `list_indices_tool` and is logged as an error. Fix should be:
```
resolve_response = await client.indices.resolve_index(name=resolve_pattern)
```

Hailong-am: thanks for catching this, updated.

nathaliellenaa (approved):
Thanks @Hailong-am , LGTM

PR #281, opened 2026-07-14
Add caller attribution to tool execution structured logs

natel926:
### Description
When different agents or applications share a single MCP server instance, there is no way to attribute tool execution metrics to a specific caller. This adds support for an optional X-MCP-Caller HTTP header that agents or applications can set to identify themselves.

Changes:
- New caller_context.py: pure-ASGI middleware extracts X-MCP-Caller header into a contextvars.ContextVar (defaults to "unknown" if absent)
- tool_executor.py: includes "caller" field in every tool_execution log event
- streaming_server.py: registers the middleware on the Starlette app
- Tests for the middleware and the updated log field

The header is optional and backward-compatible — existing clients that do not send the header will appear as caller="unknown" in logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - `ListIndexTool` now falls back to `GET /_resolve/index/*` when `_cat/indices` returns 403, allowing users with only index-level read permissions to list indices. The response is annotated when the fallback is used to indicate that health, size, and doc count are unavailable ([#279](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/279))
 - Add skills tools for root-cause analysis (`DataDistributionTool`, `LogPatternAnalysisTool`, `MetricChangeAnalysisTool`) that surface categorical value shifts, ML-clustered log patterns, and percentile changes between a baseline and an anomaly window. Skills tools are in the `skills_tools` category and can be enabled via `enabled_categories: ["skills_tools"]` ([#259](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/259))
 - Add `PPLQueryTool` for executing PPL (Piped Processing Language) queries via `/_plugins/_ppl` endpoint, with support for `jdbc`, `csv`, and `raw` output formats. Tool is in the `observability` category. ([#257](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/257))
+- Add client attribution to tool execution structured logs via optional `X-MCP-Client-Name` HTTP header. When multiple clients share a single MCP server, each client can identify itself and the `client_name` field appears in `tool_execution` log events for per-client metric filtering ([#281](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/281))
 
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


src/mcp_server_opensearch/caller_context.py
```diff
@@ -0,0 +1,56 @@
...
+        """Extract caller header and forward to the wrapped app."""
+        if scope['type'] not in ('http', 'websocket'):
+            await self.app(scope, receive, send)
+            return
+
+        # Extract X-MCP-Caller from raw ASGI headers (list of [name, value] byte pairs).
+        caller = 'unknown'
+        for header_name, header_value in scope.get('headers', []):
+            if header_name == _HEADER_NAME:
+                caller = header_value.decode('latin-1')
```
mingshl: to avoid the malicious header, better to set a length limit to make sure the header_value is certain length limit. This header is also going to add to logs better to sanitize the length

mingshl:
would you add some unit test?

natel926:
> would you add some unit test?

Unit tests were already included (test_caller_context.py + test_tool_executor.py), but I've expanded them with 8 additional tests covering the sanitization edge cases (injection, oversized headers, etc.).

src/mcp_server_opensearch/caller_context.py
natel926: Yes. we should set the length limit. Added _sanitize_caller() with a 64-char length limit and alphanumeric+hyphen+underscore+dot allowlist. Invalid values fall back to "unknown".

jiapingzeng (approved):

mingshl (approved):

nathaliellenaa:
Thanks @natel926 , overall looks good! A couple of things:

- Docs: Could you add a short section to the user guide covering the X-MCP-Caller usage?
- SSE behavior: Over SSE, tools run in the long-lived `GET /sse` task, but tool calls come in as separate `POST /messages/` requests. So X-MCP-Caller only works if set on the `/sse` connection — set it on a `POST /messages/` request and it's ignored, falling back to unknown. Is per-message attribution the intent? If not, you can document it on the user guide as well

natel926:
> Thanks @natel926 , overall looks good! A couple of things:
> 
> * Docs: Could you add a short section to the user guide covering the X-MCP-Caller usage?
> * SSE behavior: Over SSE, tools run in the long-lived `GET /sse` task, but tool calls come in as separate `POST /messages/` requests. So X-MCP-Caller only works if set on the `/sse` connection — set it on a `POST /messages/` request and it's ignored, falling back to unknown. Is per-message attribution the intent? If not, you can document it on the user guide as well

Good call — I'll add a section to USER_GUIDE.md covering X-MCP-Caller usage.
 
The CallerIdentificationMiddleware runs on all HTTP requests regardless of path or method, so it reads X-MCP-Caller from both GET /sse and POST /messages/. Clients that set the header at the HTTP client level will include it on every request automatically — no per-message setup needed. I'll add a note in the docs clarifying this. Will push an update shortly.

Hailong-am:
> X-MCP-Caller

is it actual user of calling the tool?  From the PR description, it sounds like more than a identifier for client. how about named it as X-MCP-ClientId?

natel926:
> > X-MCP-Caller
> 
> is it actual user of calling the tool? From the PR description, it sounds like more than a identifier for client. how about named it as X-MCP-ClientId?

Yes. It identifies the calling app/agent (for log attribution), not the end user. I avoided X-MCP-ClientId since client_id is already an OAuth term in the MCP SDK and would read as an auth credential. X-MCP-Caller is intentionally neutral, but happy to rename to X-MCP-Client-Name or whatever you prefer — just let me know.

Hailong-am:
> X-MCP-Client-Name

This make sense to me. @mingshl @jiapingzeng would like to get your feedback as well

nathaliellenaa (approved):
Thanks @natel926 , LGTM!

PR #287, opened 2026-07-27
Bind per-call connection credentials to the caller that supplied the URL

rithinpullela:
## Summary

In dynamic connection mode a caller can name the target cluster per call, through tool arguments or request headers. Previously the server would fill in whatever credentials it had configured for any URL it was given. This change makes a URL and the credentials used against it come from the same caller.

A caller-supplied URL no longer picks up the server's env basic auth, env mTLS client identity, ambient AWS credentials, or an unnamed profile. Both AWS paths behave the same way, whether signing directly or assuming a role through STS. A caller-named `aws_profile` remains a valid caller identity, since multi-cluster deployments legitimately pair one profile per URL, though a profile that fails to build is now an error rather than falling back to ambient credentials.

Credentials embedded in a URL are stripped from a caller-supplied URL, where the caller should send them explicitly. An operator-configured URL keeps them, since `opensearch-py` derives working basic auth from userinfo and existing configurations rely on that.

Also included:

- Connection URLs are reduced to host, port and path before use or logging. Unparseable ones are redacted, and a parse failure no longer echoes the text it failed on, which could repeat a password back to the caller.
- Agentic-memory request bodies exclude connection fields, so credentials cannot reach stored documents.
- Host classification matches on a suffix rather than a substring.
- A tool call may raise `opensearch_ssl_verify` but no longer lower it, leaving `OPENSEARCH_SSL_VERIFY` as the only way to disable certificate verification.
- An opt-in `OPENSEARCH_SSRF_GUARD` restricts caller URLs to public addresses. While enabled, such a URL does not follow redirects, since a redirect could reach an address the guard never checked, and a refused redirect is not retried through the inherited request path. The guard also rejects multicast and IPv6-wrapped IPv4 addresses, which Python treats as global. Operator-configured URLs follow redirects as before, since proxies in front of a cluster rely on that.

`OPENSEARCH_NO_AUTH` keeps its existing meaning: when it is set, the server attaches no credentials, and it remains the first branch in the authentication order. The empty-password checks yield to it, so a call is not rejected over a credential that would not have been used. One ordering wrinkle remains: the boto3 session is constructed before that branch is reached, so a call that sets `opensearch_no_auth` and also names an `aws_profile` that cannot be built fails on the profile instead of connecting anonymously. Moving the no-auth check ahead of session construction is a small follow-up.

## Observable changes for existing users

- Supplying `opensearch_url` without credentials in the same call is now an error instead of using the server's credentials. Any of basic auth, AWS keys plus region, `aws_iam_arn` together with `aws_profile`, or `opensearch_no_auth` satisfies it.
- A header `opensearch-url` is ignored in multi mode, where the registered cluster URL is authoritative.
- A false `opensearch_ssl_verify` in a tool call is ignored. Operators who want verification off set `OPENSEARCH_SSL_VERIFY=false`.

`README.md` and `USER_GUIDE.md` are updated, including a section on `OPENSEARCH_SSRF_GUARD` and what it restricts.

## Test plan

- [x] 643 unit tests pass; `ruff check` and `ruff format --check` clean
- [x] 17 new integration tests across `integration_tests/errors/test_credential_isolation.py`, `test_connection_guards.py`, and `tools/test_dynamic_connection.py`. These run a real server subprocess and MCP client, and need no cluster or AWS credentials
- [x] Each new check was mutation-tested: removing it makes its tests fail. The integration tests assert specific messages rather than that "an error occurred", so they cannot pass vacuously
- [x] Existing paths verified unbroken: env URL with env credentials, env URL with header credentials (per-user token to a fixed cluster), caller URL with caller basic auth, caller URL with `aws_iam_arn` plus `aws_profile`, caller URL with `no_auth`, operator URL carrying userinfo, and multi-mode registered clusters
- [x] Green on Linux, macOS and Windows
- [ ] CI integration suite against a live cluster

## Notes for reviewers

The commits are: the fixes, the `OPENSEARCH_SSRF_GUARD` documentation, the URL parse-error leak together with two SSRF guard gaps, then a revert of caller-header filtering. The filtering removed a capability callers have today, and a caller who sends their own `Authorization` header is supplying their own credential rather than reusing the server's, so it reads as hardening on its own merits rather than part of this change.

The dense part is the credential resolution in `_initialize_client_single_mode` and the branch conditions in `_create_opensearch_client`. Worth re-checking by hand: the interaction between an arg-supplied and a header-supplied URL, and that `forbid_ambient_fallback` covers both the IAM assume-role branch and the ambient branch.

Deliberately out of scope. The credential resolution reconstructs provenance after merging env, arg and header values, which is harder to read than deciding the source once up front; restructuring it suits a follow-up better than a rider on this change. Memory index creation still takes its target from caller-supplied values while using the server's own AWS credentials, because whether the memory tools should use caller credentials throughout, or never, is a product decision worth settling on its own. Pinning the validated address through the connection layer, which is what would close DNS rebinding under the SSRF guard, belongs with that same follow-up. The no-auth ordering noted above is a third.

Separately, and pre-existing on `main`: the OpenAPI-generated tools (`ClusterHealthTool`, `CountTool`, `MsearchTool`, `ExplainTool`) retype the inherited connection fields as strings, so a boolean like `opensearch_no_auth` fails validation on those tools, and `ListClustersTool` omits the fields entirely. That is why the new integration tests use `ListIndexTool`.


src/opensearch/client.py
```diff
@@ -143,6 +147,120 @@ async def get_opensearch_client(args: baseToolArgs) -> AsyncIterator[AsyncOpenSe
...
+        raise ConfigurationError(
+            f'Caller-supplied URL must use https when OPENSEARCH_SSRF_GUARD is enabled: '
+            f'{_scrub_url_userinfo(url)}'
+        )
+    host = parsed.hostname
+    if not host:
+        raise ConfigurationError(f'Caller-supplied URL has no host: {_scrub_url_userinfo(url)}')
+
+    try:
+        resolved = socket.getaddrinfo(host, parsed.port or 443, proto=socket.IPPROTO_TCP)
```
nathaliellenaa, line 243: With the SSRF guard on, this is a blocking, un-timed-out DNS lookup on the event loop for every caller request. Should we move it off the loop via `loop.getaddrinfo / run_in_executor` — or is the blocking acceptable here given the boto calls just below already block the loop and the guard is opt-in?

src/opensearch/helper.py
```diff
@@ -782,6 +782,17 @@ async def get_opensearch_version(args: baseToolArgs) -> Version:
...
 
+def _memory_request_body(args: baseToolArgs, *routing_fields: str) -> Dict[str, Any]:
+    """Serialize a memory tool's payload, excluding every connection field.
+
+    Excludes the inherited connection fields plus the given routing fields, so no
+    caller credential can reach the request body or be stored as a memory document.
+    Deriving the set from ``model_fields`` keeps it correct as the base model grows.
+    """
+    exclude = set(baseToolArgs.model_fields) | set(routing_fields)
+    return args.model_dump(exclude=exclude, exclude_none=True, by_alias=True)
```
nathaliellenaa, line 793: `_memory_request_body` always sets `by_alias=True`, which flips the flag for the three functions that previously used the default (`create_agentic_memory_session`, `delete_agentic_memory_by_query`, `search_agentic_memory`). Safe today since only `memory_type` is aliased and it's always excluded, but a future aliased field on these models would start serializing under its alias. Non-blocking

nathaliellenaa (commented):
Thanks @rithinpullela , overall looks good. Added couple comments

src/opensearch/client.py
```diff
@@ -143,6 +147,120 @@ async def get_opensearch_client(args: baseToolArgs) -> AsyncIterator[AsyncOpenSe
...
+    """Reject per-call connection overrides when the operator has disabled them.
+
+    Hiding the fields from the advertised schema is not enforcement, since a raw
+    client can still send them. Covers tool args only. Header-sourced connection
+    fields are gated by ``OPENSEARCH_HEADER_AUTH`` instead.
+    """
+    from mcp_server_opensearch.server_instructions import CONNECTION_OVERRIDE_FIELDS
+
+    explicit = os.getenv('OPENSEARCH_DYNAMIC_CONNECTION', '').strip().lower()
+    if explicit not in ('false', '0'):
```
arjunkumargiri: If OPENSEARCH_DYNAMIC_CONNECTION is not explicitly set then it will pass this check. Lets treat the case where  OPENSEARCH_DYNAMIC_CONNECTION is unset as false.

rithinpullela: Good catch, and it was broader than unset. We only rejected an explicit `false`, so the common case (`OPENSEARCH_URL` set, this var untouched) hid the fields from the schema but still honored them.

Now it defers to `is_dynamic_mode_enabled()`, the same predicate the schema uses, so the two can't drift:

```python
if is_dynamic_mode_enabled():
    return
```

Unset with a configured URL now rejects. Kept unset as auto-detect rather than `false` so zero-config still works.

src/opensearch/client.py
rithinpullela: Fixed. Confirmed it blocks: with a 1.5s slow resolver a heartbeat coroutine ticked once, 16 after the fix.

Offloaded the whole init rather than just the DNS call, so the boto3 work you mentioned moves off the loop too:

```python
client = await asyncio.to_thread(initialize_client, args)
```

src/opensearch/helper.py
rithinpullela: Looked into reverting those three to the default, but we can't: `by_alias=True` is required. Message content has to go out as `type`, and turning the flag off would send `content_type` instead, breaking `add_agentic_memories` and `update_agentic_memory`.

So the flag stays. I checked all five bodies with and without it and there's no difference today, matching what you found.

For the future-field risk you raised, added a test that fails if any body field is aliased. Verified by adding one.

src/opensearch/client.py
arjunkumargiri, line 328: Nit: 
* Python documentation of this function needs to be updated to match the new logic.
*` _initialize_client_single_mode` has become complex/big making it harder for readability. Let's take a follow up task to refactor this function into smaller functions.

arjunkumargiri (approved):

nathaliellenaa (approved):
Thanks @rithinpullela , LGTM

Issue #265, opened 2026-06-03
[FEATURE] Set CallToolResult.isError=true when a tool fails (currently always false)


Ferioney:
### Is your feature request related to a problem?
Tools should signal failures through the MCP [`CallToolResult.isError`](https://modelcontextprotocol.io/specification/draft/schema#toolresultcontent-iserror) field so clients can programmatically detect that something went wrong. Today every response comes back with `isError: false`.
For example, when a tool fails — e.g. an authentication problem on `ClusterHealthTool` — the error text is returned but `isError` is `false`:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "[{\"text\":\"Error executing ClusterHealthTool: AuthenticationException(401, 'Unauthorized')\"}]"
      }
    ],
    "isError": false
  }
}
```

### What solution would you like?
The same failure should set `isError: true`:
```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "Error executing ClusterHealthTool: AuthenticationException(401, 'Unauthorized')"
      }
    ],
    "isError": true
  }
}
```

rithinpullela:
Closing the issue since PR is merged

PR #292, opened 2026-07-29
Migrate to mcp 2.0

Hailong-am:
### Description
Upgrades the `mcp[cli]` dependency from `>=1.9.4,<2` to `>=2.0.0` and ports the server and integration test client to the mcp 2.0 API.

mcp 2.0 removes the `@server.list_tools()` / `@server.call_tool()` decorator pattern and the `request_ctx` contextvar, replacing them with constructor-injected handlers that receive request context explicitly.

Changes:
- `pyproject.toml`: remove `<2` cap and stale comment
- `stdio_server.py` / `streaming_server.py`: replace decorator registrations with `Server(on_list_tools=..., on_call_tool=...)` constructor kwargs; handler signatures now receive `(ctx, params: CallToolRequestParams)`
- `client_context.py`: add `request_context_var` ContextVar to carry the Starlette `Request` through tool calls, replacing the removed `request_ctx`
- `client.py`: switch import to `request_context_var`; `_get_auth_from_headers()` reads the `Request` directly
- `tool_executor.py`: `CallToolResult(isError=)` → `CallToolResult(is_error=)` per renamed field in mcp_types 2.0
- `integration_tests/framework/client.py`: fix import (`streamablehttp_client` → `streamable_http_client`); pass headers/timeout via `httpx2.AsyncClient` since the new function no longer accepts those params directly
- `CHANGELOG.md`: document the upgrade
- Tests: update patch targets, mock setup, and attribute assertions to match renamed fields

### Issues Resolved
_None — this upgrade was tracked in the `<2` cap comment added in #288._


src/mcp_server_opensearch/streaming_server.py
```diff
@@ -61,24 +59,35 @@ async def create_mcp_server(
...
+    async def _call_tool(ctx, params: CallToolRequestParams) -> CallToolResult:
+        from mcp_server_opensearch.client_context import request_context_var
         from mcp_server_opensearch.tool_executor import execute_tool
 
-        return await execute_tool(name, arguments, enabled_tools)
+        token = request_context_var.set(ctx.request)
+        try:
+            return await execute_tool(params.name, params.arguments or {}, enabled_tools)
+        finally:
+            request_context_var.reset(token)
```
rithinpullela, line 95: This changes error behavior vs v1. The old `@server.call_tool()` decorator caught what `execute_tool` raised and returned it as `CallToolResult(isError=True);` calling `execute_tool` directly means unknown tool / missing opensearch_cluster_name / validation errors now surface as MCPError. That breaks the #265 `isError` contract our metrics use, and on modern HTTP the message gets scrubbed to a generic "Internal server error". 


Suggest mirroring the SDK's own handler pattern:

```
try:
    return await execute_tool(params.name, params.arguments or {}, enabled_tools)
except MCPError:
    raise
except Exception as e:
    return _build_call_tool_result([str(e)], is_error=True)
finally:
    request_context_var.reset(token)
```

integration_tests/framework/client.py
```diff
@@ -24,11 +25,14 @@ async def mcp_client(server_url: str, headers: dict | None = None, timeout: floa
     """
     stack = AsyncExitStack()
     try:
-        transport_ctx = streamablehttp_client(
-            url=server_url,
+        http_client = httpx2.AsyncClient(
             headers=headers or {},
             timeout=timeout,
         )
```
rithinpullela: `httpx2.AsyncClient `is never closed — `streamable_http_client` only closes a client it created itself, so this leaks per session. Enter it on the stack like the others:

```
http_client = await stack.enter_async_context(
    httpx2.AsyncClient(headers=headers or {}, timeout=timeout)
)
```

rithinpullela (commented):
Thanks for the quick PR @Hailong-am added 2 small comments.

src/mcp_server_opensearch/streaming_server.py
Hailong-am: Thanks @rithinpullela, updated

integration_tests/framework/client.py
Hailong-am: done

src/mcp_server_opensearch/stdio_server.py
```diff
@@ -53,24 +51,39 @@ async def serve(
...
+        from mcp_server_opensearch.tool_executor import _build_call_tool_result, execute_tool
 
-        return await execute_tool(name, arguments, enabled_tools)
+        token = request_context_var.set(ctx.request)
+        try:
+            return await execute_tool(params.name, params.arguments or {}, enabled_tools)
+        except MCPError:
+            raise
+        except Exception as e:
+            return _build_call_tool_result([str(e)], is_error=True)
```
nathaliellenaa: `build_call_tool_result` only wraps dict items into `TextContent`; a bare string is appended as-is, and `CallToolResult(content=[<str>])` fails Pydantic validation. Verified directly:
```
  _build_call_tool_result(['some error string'], is_error=True)
  → ValidationError: content.0 Input should be a valid dictionary or instance of TextContent
```

So unexpected exception in a tool call hits this fallback and the fallback itself throws. The escaping `ValidationError` then propagates to the framework, which maps `ValidationError` → `INVALID_PARAMS`, the real tool error text is lost and the client gets an opaque protocol error.

Suggested fix: pass a dict, matching the `log_tool_error` shape used everywhere else:
```
return _build_call_tool_result([{'type': 'text', 'text': str(e)}], is_error=True)
```
Same fix is needed in streaming_server.py

Hailong-am: good catch, updated and add few unit test cases.

Hailong-am:
@nathaliellenaa @rithinpullela could you help to review it again?

rithinpullela (approved):

PR #301, opened 2026-08-15
Expose tool category in _meta field of tools/list response

### LINKED ISSUE #300: Expose tool category in tools/list _meta field
## Problem

The \`tools/list\` MCP response currently returns only \`name\`, \`description\`, and \`inputSchema\` for each tool. Clients have no way to know which category a tool belongs to (\`core_tools\`, \`search_relevance\`, \`observability\`, etc.) without parsing tool names heuristically.

## Solution

Expose the category in the \`_meta\` field of each \`Tool\` object in the \`tools/list\` response:

\`\`\`json
{
  "name": "ListIndexTool",
  "description": "...",
  "inputSchema": { ... },
  "_meta": { "category": "core_tools" }
}
\`\`\`

This is safe for all existing clients. The MCP 1.x (2024-11-05) JSON schema for \`Tool\` does not set \`additionalProperties: false\`, so extra properties are valid by spec — raw HTTP clients validating against the schema will accept the response. Clients using the Python SDK will also accept it since Pydantic v2 silently drops unknown fields by default.

## Changes

- Extract \`BUILTIN_CATEGORY_TOOLS\` as a module-level constant — single source of truth for category → tool mappings, replacing ~100 lines of inline list-building in \`process_tool_filter\`
- Add \`build_category_map()\` helper
- \`process_tool_filter\` returns \`category_to_tools\`; \`get_tools\` stamps \`tool_info['category']\` on every enabled tool in both single and multi mode
- \`stdio_server\` and \`streaming_server\` pass \`meta={'category': ...}\` to the \`Tool\` constructor; tools with no known category emit \`meta=None\`
- Unit tests for \`build_category_map\`, return value of \`process_tool_filter\`, and category stamping
- Integration test verifying core tools carry correct \`_meta.category\` against a live server

Hailong-am:
## Summary

Closes #300

- Extract `BUILTIN_CATEGORY_TOOLS` as a module-level constant in `tool_filter.py` — single source of truth for category → tool mappings, replacing ~100 lines of inline list-building scattered inside `process_tool_filter`
- Add `build_category_map()` helper that translates registry keys to display names
- `process_tool_filter` now returns `category_to_tools`; `get_tools` uses it to stamp `tool_info['category']` on every enabled tool in both single and multi mode
- `stdio_server` and `streaming_server` pass `meta={'category': ...}` to the `Tool` constructor so category is serialised as `_meta` on the wire; tools with no known category emit `meta=None`

**Compatibility note:** The MCP 1.x (2024-11-05) JSON schema for `Tool` does not set `additionalProperties: false`, so extra properties are valid by spec for all existing clients including raw HTTP ones.

## Test plan

- [ ] Unit tests: `TestBuildCategoryMap` — mapping correctness, empty registry, custom display names
- [ ] Unit tests: `TestCategoryStamping` — `get_tools` stamps category in single and multi mode; unknown tools get `''`
- [ ] Integration test: `integration_tests/server_modes/test_list_tools.py` — core tools carry `_meta.category = "core_tools"` against a live server; all listed tools have a known category or no `_meta`
- [ ] Manual: start server against local OpenSearch, run `tools/list` via curl and confirm `_meta.category` appears in the response

### DIFF
--- a/src/tools/tool_filter.py
+++ b/src/tools/tool_filter.py
@@ -132,6 +132,22 @@ def process_categories(category_list, category_to_tools):
 
 # Single source of truth: built-in category → registry key lists.
 # Used by build_category_map and to stamp tool_info['category'] at startup.
+#
+# 'observability' and 'skills' are fine-grained categories that existed before
+# the unified 'analytics' category was introduced.  'analytics' is a superset
+# that contains every tool from both.  All three are first-class categories —
+# enabling 'skills' gives the 3 skills tools, enabling 'observability' gives
+# PPLQueryTool, and enabling 'analytics' gives all 4.
+_SKILLS_TOOLS: list[str] = [
+    'DataDistributionTool',
+    'LogPatternAnalysisTool',
+    'MetricChangeAnalysisTool',
+]
+
+_OBSERVABILITY_TOOLS: list[str] = [
+    'PPLQueryTool',
+]
+
 BUILTIN_CATEGORY_TOOLS: dict[str, list[str]] = {
     'core_tools': [
         'ListIndexTool',
@@ -179,32 +195,10 @@ BUILTIN_CATEGORY_TOOLS: dict[str, list[str]] = {
         'DeleteAgenticMemoryByQueryTool',
         'SearchAgenticMemoryTool',
     ],
-    'analytics': [
-        'PPLQueryTool',
-        'DataDistributionTool',
-        'LogPatternAnalysisTool',
-        'MetricChangeAnalysisTool',
-    ],
+    'observability': _OBSERVABILITY_TOOLS,
+    'skills': _SKILLS_TOOLS,
+    'analytics': _OBSERVABILITY_TOOLS + _SKILLS_TOOLS,
 }
-
-
-# Aliases allow legacy category names to resolve to their replacement.
-# Enabling 'observability' or 'skills' expands to 'analytics'.
-CATEGORY_ALIASES: dict[str, list[str]] = {
-    'observability': ['analytics'],
-    'skills': ['analytics'],
-}
-
-
-def expand_category_aliases(categories: list[str]) -> list[str]:
-    """Expand any alias names in a category list to their constituent categories."""
-    expanded = []
-    for cat in categories:
-        if cat in CATEGORY_ALIASES:
-            expanded.extend(CATEGORY_ALIASES[cat])
-        else:
-            expanded.append(cat)
-    return expanded
 
 
 def build_category_map(tool_registry: dict) -> dict[str, list[str]]:
@@ -333,10 +327,6 @@ def process_tool_filter(
                     f'Tools exempt from write filter via allow_write_categories: {exempt_tools}'
                 )
             apply_write_filter(tool_registry, exempt_tools=exempt_tools)
-
-        # Expand category aliases (e.g. 'observability'/'skills' -> 'analytics')
-        enabled_category_list = expand_category_aliases(enabled_category_list)
-        disabled_category_list = expand_category_aliases(disabled_category_list)
 
         # Process tools from categories and regex patterns
         enabled_tools_from_categories = process_categories(
--- a/src/tools/tool_filter.py
+++ b/src/tools/tool_filter.py
@@ -5,7 +5,6 @@ import json
 import logging
 import os
 import re
-from .skills_tools import SKILLS_TOOLS_REGISTRY
 from .tool_params import baseToolArgs
 from .utils import (
     is_tool_compatible,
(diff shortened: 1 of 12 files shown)


Hailong-am:
@nathaliellenaa @rithinpullela @mingshl could you help to review this PR to add tool category returned in tool list call?

src/tools/tool_filter.py
```diff
@@ -131,6 +130,80 @@ def process_categories(category_list, category_to_tools):
...
+        'GetAgenticMemoryTool',
+        'UpdateAgenticMemoryTool',
+        'DeleteAgenticMemoryByIDTool',
+        'DeleteAgenticMemoryByQueryTool',
+        'SearchAgenticMemoryTool',
+    ],
+    'observability': [
+        'PPLQueryTool',
+    ],
+    'skills': [
```
mingshl: Nice to see all tool filters defined in mcp, then on the agent side, it would be able to call the tools with these well define filters. But maybe a nit comment on the tool filter name called "skills" would be a bit confusing. Usually this filter would be used in codes, but dynamic selected by LLM, then probably this filter name doesn't matter.

mingshl (approved):

src/tools/tool_filter.py
Hailong-am: This category metadata won't exposed to LLM and it only used in code. `Analytics` might be a better name.

src/tools/tool_filter.py
```diff
@@ -131,6 +130,97 @@ def process_categories(category_list, category_to_tools):
...
+    ],
+}
+
+
+# Aliases allow legacy category names to resolve to their replacement.
+# Enabling 'observability' or 'skills' expands to 'analytics'.
+CATEGORY_ALIASES: dict[str, list[str]] = {
+    'observability': ['analytics'],
+    'skills': ['analytics'],
+}
```
rithinpullela, line 201: **The category merge changes what existing configs enable, but no shipped doc or changelog line says so.**

Concretely (verified by executing the filter on this branch):
- `OPENSEARCH_ENABLED_CATEGORIES=observability` previously enabled 1 tool (`PPLQueryTool`); it now enables 4.
- `OPENSEARCH_ENABLED_CATEGORIES=skills` previously enabled 3 tools; it now also pulls in `PPLQueryTool`.

All four are read-only analysis/query tools, so there's no write-exposure risk — but README still documents Observability and Skills as separate categories with different tool lists, the server-instructions docstring (`server_instructions.py:169-170`) still says only `skills`, and the CHANGELOG entry describes only the `_meta` feature, not the category merge.

The merge itself is a reasonable design — this just needs the docs to land in this PR:
1. Replace the two README sections with one Analytics section listing all four tools, noting `observability` and `skills` are compatibility aliases that each enable the full merged set.
2. Update the USER_GUIDE tool-filter section the same way.
3. Update the `server_instructions.py` docstring to name `analytics` as canonical.
4. Add a CHANGELOG line for the merge itself (e.g., "Merged `observability` and `skills` categories into `analytics`; old names remain as aliases enabling the merged set").
5. Add tests for both alias names in enabled *and* disabled lists (currently only enabled-`skills` is tested, `test_tool_filters.py:616-641`).

src/tools/tool_filter.py
```diff
@@ -349,6 +334,10 @@ def process_tool_filter(
                 )
             apply_write_filter(tool_registry, exempt_tools=exempt_tools)
 
+        # Expand category aliases (e.g. 'observability'/'skills' -> 'analytics')
+        enabled_category_list = expand_category_aliases(enabled_category_list)
+        disabled_category_list = expand_category_aliases(disabled_category_list)
```
rithinpullela: **User-defined categories named `skills` or `observability` are silently clobbered by alias expansion — regression introduced by this PR.**

Custom categories merge into `category_to_tools` at line 277, but `expand_category_aliases` rewrites those names in enabled/disabled lists unconditionally — it only consults `CATEGORY_ALIASES`, never checking whether the user defined a category with that name.

Verified at both commits with the same setup (`tool_category: {skills: [CustomTool]}` + `enabled_categories: [skills]`):
- At merge-base `9b20641`: `CustomTool` is enabled — the custom definition is honored.
- At this PR's head: `CustomTool` is silently dropped and the analytics tools are enabled instead.

Suggested fix: in `expand_category_aliases`, skip expansion for any name the user explicitly defined in `category_to_tools` (explicit definition wins over alias) — or reject the reserved names at config load with a clear startup error. Either is a few lines.

src/tools/tool_filter.py
```diff
@@ -131,6 +130,97 @@ def process_categories(category_list, category_to_tools):
...
+        'DeleteAgenticMemoryByIDTool',
+        'DeleteAgenticMemoryByQueryTool',
+        'SearchAgenticMemoryTool',
+    ],
+    'analytics': [
+        'PPLQueryTool',
+        'DataDistributionTool',
+        'LogPatternAnalysisTool',
+        'MetricChangeAnalysisTool',
+    ],
```
rithinpullela: **Hardcoding the analytics tool list re-introduces drift the old code prevented.**

Before this PR, the `skills` category was built by iterating `SKILLS_TOOLS_REGISTRY`, so a new skills tool automatically got category membership. This PR repeats the three key strings manually — the next tool added to `SKILLS_TOOLS_REGISTRY` (`skills_tools.py:236`, still spread into `TOOL_REGISTRY` at `tools.py:981`) will register successfully yet be invisible under `analytics`, unfilterable by category, and carry no `_meta.category`. That's a second source of truth in a PR whose description says "single source of truth."

Suggested fix (one line, restoring the import this PR removed):

```python
'analytics': ['PPLQueryTool', *SKILLS_TOOLS_REGISTRY.keys()],
```

Alternatively, keep the literal list but add an invariant test asserting every `SKILLS_TOOLS_REGISTRY` key appears in `BUILTIN_CATEGORY_TOOLS['analytics']` (the empty-registry test at `test_tool_filters.py:86-90` deliberately skips analytics, so nothing enforces this today).

rithinpullela (commented):
Reviewed at commit 6fca889. The `_meta` mechanism itself checks out — I verified against the pinned mcp 2.0.0 SDK that `Tool(meta=...)` serializes as `_meta` and `meta=None` omits the field entirely, and the refactor into `BUILTIN_CATEGORY_TOOLS`/`build_category_map()` is behavior-preserving for the four untouched categories (exact set-diff, zero changes). Three comments below on the category-merge side of the change. Also: this branch predates #302 but test-merges cleanly with latest main (only CHANGELOG overlaps), so no rebase needed.

src/tools/tool_filter.py
Hailong-am: Addressed in 69bf9e4. Instead of merging into a single `analytics` category with aliases, I kept all three as independent first-class categories:

- `observability` → PPLQueryTool (1 tool)
- `skills` → DataDistributionTool, LogPatternAnalysisTool, MetricChangeAnalysisTool (3 tools)
- `analytics` → superset of both (all 4 tools)

No alias expansion, so `OPENSEARCH_ENABLED_CATEGORIES=skills` enables exactly the 3 skills tools — no behavior change from the base branch.

Docs updated: README (both sections mention `analytics` superset), USER_GUIDE (added Built-in Categories table), DEVELOPER_GUIDE (updated category list), CHANGELOG (added entry for the new `analytics` category).

src/tools/tool_filter.py
Hailong-am: Fixed in 69bf9e4. `CATEGORY_ALIASES` and `expand_category_aliases` are removed entirely. `observability` and `skills` are now real categories in `BUILTIN_CATEGORY_TOOLS`, so a user-defined `tool_category: {skills: [CustomTool]}` with `enabled_categories: [skills]` will merge the user's custom tools on top of the built-in list (the existing `category_to_tools.update()` behavior) — no silent clobbering.

src/tools/tool_filter.py
Hailong-am: Addressed in 69bf9e4 with two invariant tests:

1. `test_skills_category_matches_skills_tools_registry` — asserts `BUILTIN_CATEGORY_TOOLS['skills']` == `SKILLS_TOOLS_REGISTRY.keys()`. Fails if someone adds a tool to `skills_tools.py` but forgets `_SKILLS_TOOLS`.
2. `test_analytics_is_superset_of_observability_and_skills` — asserts `analytics` == `observability ∪ skills`. Fails if a sub-category grows but the superset doesn't.

I kept the explicit list (`_SKILLS_TOOLS`) rather than `*SKILLS_TOOLS_REGISTRY.keys()` because the import was removed in the refactor and the invariant test provides the same safety net. Happy to switch to the import if you prefer.

rithinpullela (approved):

PR #306, opened 2026-09-01
Support multiple datasources via headers

jiapingzeng:
### Description

Adds support for targeting multiple OpenSearch datasources within a single request when the server runs in **multi mode with header auth** (`--mode multi` / `OPENSEARCH_MODE=multi` + `OPENSEARCH_HEADER_AUTH=true`).

A request carries the datasources as aligned, comma-separated headers:

- `opensearch-url`: `url0,url1,...`
- `opensearch-cluster-name`: `name0,name1,...` — optional; defaults to `cluster1,cluster2,...`
- `aws-service-name`: `es,aoss,...`
- `aws-region`: `region0,region1,...`

All provided lists must align 1:1 (one entry per datasource). The credential headers (`aws-access-key-id`/`aws-secret-access-key`/`aws-session-token`, or `Authorization`) remain scalar and are shared across all datasources.

The LLM never sees a URL: it calls `ListClustersTool` to discover the datasource names for the request, then passes the chosen name as the `opensearch_cluster_name` tool argument. The server maps that name to its URL, region, and service and signs with the shared credential.

Header URLs are treated as caller-supplied: they get the same SSRF-guard and ambient-credential protections as single-mode header auth.

### Breaking changes (multi mode + header auth)

- Per-datasource `aws-region` / `aws-service-name` now come from the resolved datasource, not raw request headers. A YAML cluster with `opensearch_header_auth: true` that relied on those headers must set region/serverless in config or `AWS_REGION` (else region errors, service defaults to `es`).
- `aws-service-name` / `aws-region` must align 1:1 with `opensearch-url` — a single value is no longer broadcast to all datasources. Credential headers stay shared/scalar.

### Issues Resolved

N/A

### Issues Resolved


vamsimanohar:
### The same credential goes to all the datasources

#### The problem

This PR sends one credential to every datasource. With two datasources on different hosts, the credential of `alpha` goes to the host of `beta`:

```
BASIC  select=beta  -> https://beta.example.com:443   auth=('alpha-user','alpha-secret')
BEARER select=beta  -> https://beta.example.com:443   sent={'Authorization': 'Bearer TOKEN-FOR-ALPHA'}
```

The model selects that host at the time of the tool call.

The data on the wire is not the same for each type of credential. With SigV4 the server sends a signature. The host, the region and the service are in the signed string, and the secret key stays on the server. So a signature for domain A does not work against domain B. With a bearer token or a password, the server sends the secret itself. The host can then use that secret again.

The docstring of `_ambient_aws_fallback_allowed` makes the same difference: *"Basic auth, bearer tokens, and mTLS certs are sent to the named host verbatim, so they are never shared."* The new code does not keep this rule.

One credential also stops a valid use case. A request cannot use a different credential for each datasource. These configurations are not possible now:

- Two clusters with different basic-auth users.
- One AWS domain with SigV4 and one self-managed cluster with basic auth.
- Two domains in different AWS accounts, with a different key for each account.

#### The solution

Give each datasource its own credential. A scalar header cannot do this, so use one header with a JSON value. The structure is the same as the YAML `clusters` block, and `ClusterInfo` stays the only schema:

```
opensearch-datasources: {"prod":{"opensearch_url":"https://a…","aws_region":"us-east-1"},
                         "stage":{"opensearch_url":"https://b…","opensearch_username":"u","opensearch_password":"p"}}
```

Read this header when `OPENSEARCH_HEADER_AUTH` is `true`. Read it in single mode and in multi mode. In single mode the header must have one entry. In multi mode the names are the JSON keys, and `ListClustersTool` gives them to the model.

Then apply this rule:

- AWS keys: the request can share them between all the datasources. One IAM identity often has access to more than one domain. Each datasource keeps its own region and service, so each request has the correct signature.
- Bearer token or basic auth: each entry must hold its own credential. If an entry holds no credential, use the `Authorization` header only when the request has one datasource. This keeps the present behaviour of a deployment with header auth and one domain, and it makes the unsafe condition impossible.

This format gives both results. It stops the unsafe condition, and it makes a different credential for each datasource possible.

For a small change in this PR, you can correct the safety problem only. Reject a request with more than one datasource if the credential is a bearer token or basic auth. The JSON header can come later.

jiapingzeng:
> This PR sends one credential to every datasource. With two datasources on different hosts, the credential of alpha goes to the host of beta

This is intentional. Passing multiple sets of credentials via headers is not scalable as fields such as aws session token can be very long and exceed header size limit.

At the moment the use case of this feature is a single user (with single set of credentials) accessing multiple data sources that they have access to.

mingshl:
@jiapingzeng and @vamsimanohar I do align on one credentials for all data sources is a valid concern, I have another proposal using indexed header in the PR https://github.com/opensearch-project/opensearch-mcp-server-py/pull/309

src/opensearch/client.py
```diff
@@ -1212,3 +1226,137 @@ def _get_auth_from_headers() -> Dict[str, Optional[str]]:
...
+
+def _datasource_names(urls: list[str], header_auth: Dict[str, Optional[str]]) -> list[str]:
+    """Datasource names from the opensearch-cluster-name header, or generated cluster1..clusterN.
+
+    The name header is optional; when omitted we synthesize placeholder names so a client that
+    sends only the routing headers still works.
+    """
+    names = _split_header_list(header_auth.get('cluster_names'))
+    if not names:
+        names = [f'cluster{i + 1}' for i in range(len(urls))]
```
nathaliellenaa: If `opensearch-cluster-name` is provided, the comma count check works as expected and catches a mismatch. But when it's omitted, we generate the names from `len(urls)` here — so isn't the count check comparing the URL count to itself, meaning it can never fail? Should we require `opensearch-cluster-name` whenever `opensearch-url` has more than one entry?

src/opensearch/client.py
```diff
@@ -89,15 +89,24 @@ def initialize_client(args: baseToolArgs) -> AsyncOpenSearch:
...
-            # Get cluster information
-            cluster_info = get_cluster(args.opensearch_cluster_name)
-            if not cluster_info:
-                raise ConfigurationError(
-                    f'Cluster "{args.opensearch_cluster_name}" not found in configuration'
+            # With header auth, datasources are defined by request headers (mutually
+            # exclusive with the YAML registry); otherwise use the registry.
+            from mcp_server_opensearch.server_instructions import is_header_auth_enabled
+
+            if is_header_auth_enabled():
```
nathaliellenaa, line 96: With this approach, when `OPENSEARCH_HEADER_AUTH=true` in multi mode, `get_cluster()` is never called — so the YAML registry becomes unreachable and registered cluster names stop resolving. Is that intended?

USER_GUIDE.md
```diff
@@ -404,6 +404,15 @@ For Bearer authentication:
...
 
+**Multiple datasources per request (multi mode):** With `--mode multi` (or `OPENSEARCH_MODE=multi`) and `OPENSEARCH_HEADER_AUTH=true`, a single request can carry several datasources as comma-separated, positionally aligned headers:
+
+- `opensearch-url`: `url0,url1,...`
+- `opensearch-cluster-name`: `name0,name1,...` — the LLM-facing selector (optional; defaults to `cluster1,cluster2,...` when omitted)
+- `aws-service-name`: `es,aoss,...`
+- `aws-region`: `region0,region1,...`
+
+All provided lists must align 1:1 (one entry per datasource) — a single value is not broadcast to multiple datasources. The credential headers (`aws-access-key-id`/`aws-secret-access-key`/`aws-session-token`, or `Authorization`) stay scalar and are shared across all datasources. The LLM calls `ListClustersTool` to discover the available names, then passes the chosen name as `opensearch_cluster_name` on each tool call; the server maps that name to its URL, region, and service. Header-defined datasources take precedence over the YAML cluster registry for that request.
```
nathaliellenaa: > stay scalar and are shared across all datasources.

Nit: The shared-credential behavior is buried mid-paragraph between the alignment rule and the selection flow. Can we pull it out as a standalone `Note:` right after the header bullet list (~line 412) so it's more visible?

src/opensearch/client.py
jiapingzeng: The intention of this generated naming is for backwards compatibility where the frontend will send requests without the cluster name header. I think your suggestion of requiring names when there are more than one urls makes sense as older frontend client will only send a single data source, will take a look there.

src/opensearch/client.py
jiapingzeng: Yes, when both header auth and multi mode are enabled, only headers will be used and we do not fall back to yaml. In header auth mode the server should not try to read local files. Added as a potential breaking change in changelog.

nathaliellenaa (approved):
Thanks @jiapingzeng , LGTM

PR #309, opened 2026-09-09
[DRAFT] Multi-datasource support with indexed headers (alternative to PR #306)

mingshl:
## 🎯 Overview

This PR implements multi-datasource support using **indexed HTTP headers** as an alternative to the comma-separated approach in PR #306.

## 🔒 Security Advantages Over PR #306

This design addresses **three critical security issues** identified in PR #306:

### 1. Credential Confusion Vulnerability

**Problem in PR #306** ([vamsimanohar's comment](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/306#issuecomment-5591959562)):
```http
# Single credential goes to ALL datasources
opensearch-url: https://alpha.com,https://beta.com
Authorization: Bearer TOKEN-FOR-ALPHA

# When LLM selects beta:
❌ Request goes to beta.com with alpha's credential!
❌ Basic Auth/Bearer tokens leak to wrong host
❌ beta can replay alpha's credential
```

**Our Solution:**
```http
# Per-datasource credentials
opensearch-url-0: https://alpha.com
aws-access-key-id-0: AKIA...ALPHA
aws-secret-access-key-0: secret-alpha

opensearch-url-1: https://beta.com
opensearch-username-1: beta-user
opensearch-password-1: beta-pass
```

Each datasource has isolated credentials. No cross-contamination possible.

### 2. Delimiter Injection Risk

**Problem in PR #306** (AutoSDE warning on CR-302018774 line 187):

> If any title (or endpoint) contains a comma, the joined lists desynchronize and the MCP server will pair a datasource name with the wrong endpoint — silently routing a query to a different cluster than the one selected (cross-datasource exposure).

**Our Solution:**
- No delimiters → no injection risk
- Each value in its own HTTP header
- Cannot desynchronize by design

### 3. Single Point of Failure

**Problem in PR #306:**
```json
{"alpha": {...}, "beta": {SYNTAX_ERROR "gamma": {...}}
```
One malformed JSON → entire request fails

**Our Solution:**
```http
opensearch-url-0: https://alpha.com  ✅
opensearch-url-1: [CORRUPTED]       ❌
opensearch-url-2: https://gamma.com  ✅
```
Graceful degradation: datasources 0 and 2 still work

## 🏗️ Design Comparison

| Feature | Indexed Headers (This PR) | Comma-Separated (PR #306) |
|---------|---------------------------|---------------------------|
| **Credential Isolation** | ✅ Per-datasource | ❌ Shared (vulnerability) |
| **Delimiter Injection** | ✅ None | ❌ AutoSDE warning |
| **Graceful Degradation** | ✅ Single header failure isolated | ❌ JSON parse failure = total failure |
| **Debuggability** | ✅ grep-able logs | ⚠️ Requires JSON parsing |
| **HTTP Native** | ✅ Standard headers | ✅ Standard headers |
| **Header Size** | ✅ Each ~8KB limit | ⚠️ Single header may exceed |
| **Parsing Complexity** | ✅ Simple while loop | ⚠️ split() + alignment validation |

### Header Format

```http
# Multiple datasources (indexed)
opensearch-url-0: https://domain1.us-west-2.es.amazonaws.com
opensearch-url-1: https://collection1.us-east-1.aoss.amazonaws.com
opensearch-url-2: https://domain2.eu-west-1.es.amazonaws.com

# Credentials per datasource (prevents confusion attack)
aws-access-key-id-0: AKIA...
aws-secret-access-key-0: secret1
aws-region-0: us-west-2

opensearch-username-1: user2
opensearch-password-1: pass2
aws-region-1: us-east-1

# Backward compatible: single datasource still works
opensearch-url: https://domain1.com
aws-access-key-id: AKIA...
```

### Key Functions

```python
✅ _get_allowed_datasources_from_headers()  # Extract allowlist with gap detection
✅ _normalize_opensearch_url()              # URL normalization
✅ _validate_opensearch_url()               # Fail-closed validation
```

### Security Properties

1. ✅ **Fail Closed** - Rejects requests on validation failure (no fallback)
2. ✅ **Allowlist Only** - Only URLs from gateway headers are permitted
3. ✅ **Immutable** - LLM cannot modify allowed datasources
4. ✅ **Normalized Comparison** - Prevents bypass via case/trailing-slash
5. ✅ **Per-datasource Credentials** - No credential confusion (fixes vamsimanohar's concern)
6. ✅ **Sparse Index Detection** - Warns on gaps (e.g., 0, 2, 3 missing 1)

## 🧪 Testing

```
✅ 17 test cases covering:
   - URL normalization (6 tests)
   - Validation success (5 tests)
   - Validation failure (4 tests)
   - Multi-datasource flow (1 test)
   - Backward compatibility (1 test)

All tests passing: 5/5 test suites
```

## 🔬 Design Analysis

We conducted a detailed comparison of three approaches:
1. **Indexed headers (this PR)** - Score: 38/50
2. Comma-separated headers (PR #306) - Score: 31/50
3. JSON header (vamsimanohar's proposal) - Score: 33/50

**Key findings favoring indexed headers:**
- **Operational Excellence**: 5/5 debuggability vs 2/5 for others (grep-able logs)
- **Edge Case Resilience**: Graceful degradation vs catastrophic failure
- **Amazon Infrastructure Fit**: AWS tooling assumes one-value-per-header
- **Zero-Downtime Migration**: Perfect backward compatibility

**Trade-offs accepted:**
- ❌ No semantic datasource names (get indices 0,1,2 not "prod","stage")
  → Can be added later via supplementary header if needed
- ❌ Harder to extend with metadata
  → Not needed for MVP, can evolve protocol

## 🔗 Related

- **Alternative approach**: PR #306 (comma-separated)
- **Security discussions**:
  - [vamsimanohar's credential confusion concern](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/306#issuecomment-5591959562)
  - AutoSDE delimiter injection warning (Amazon internal CR)

## 📋 Checklist

- [x] Implementation complete
- [x] Tests added and passing (17/17)
- [x] Backward compatible
- [x] Security validated (fail-closed)
- [x] Per-datasource credentials (no confusion)
- [x] No delimiter injection risk
- [x] Sparse index detection added
- [ ] Code review
- [ ] Integration testing with gateway

## 💬 Discussion Points

This is a **draft PR** to present an alternative design. Key questions for the community:

1. **Security vs Convenience**: This design prioritizes security (per-datasource credentials, no injection) over convenience (no semantic names). Is this the right trade-off?

2. **Debuggability**: Indexed headers are grep-able and human-readable in logs. Is this important for your operational model?

3. **Extensibility**: If semantic names are essential, we could add supplementary headers (e.g., `opensearch-datasource-name-0: prod`). Would this address concerns?

4. **Credential isolation**: Do you need to support mixed credentials (SigV4 for cluster A, Basic Auth for cluster B) in a single request?

Looking forward to feedback from @jiapingzeng, @vamsimanohar, and the community!

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 
 ### Added
+- Add multi-datasource support with indexed headers (`opensearch-url-0`, `opensearch-url-1`, ...) and security validation. The MCP server now validates requested datasource URLs against an allowlist extracted from gateway headers, preventing LLM agents from accessing unauthorized datasources. Supports per-datasource credentials and graceful degradation. Backward compatible with single-datasource (non-indexed) headers. ([#309](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/309))
 - Expose tool category in the `_meta` field of each `Tool` object returned by `tools/list`. Core tools report `"_meta": {"category": "core_tools"}`; tools with no known category omit `_meta`. The MCP 1.x schema for `Tool` does not restrict additional properties, so this is safe for all existing clients. ([#301](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/301))
 - Add `analytics` category as a superset of `observability` (PPLQueryTool) and `skills` (DataDistributionTool, LogPatternAnalysisTool, MetricChangeAnalysisTool). All three categories are independent — `enabled_categories=skills` enables the 3 skills tools, `enabled_categories=observability` enables PPLQueryTool, and `enabled_categories=analytics` enables all 4. ([#301](https://github.com/opensearch-project/opensearch-mcp-server-py/pull/301))
 
--- a/src/opensearch/client.py
+++ b/src/opensearch/client.py
@@ -532,6 +532,11 @@ def _initialize_client_single_mode(args: baseToolArgs = None) -> AsyncOpenSearch
         if caller_supplied_url or header_supplied_url:
             _reject_caller_url_if_not_public(opensearch_url)
 
+        # SECURITY: Validate opensearch_url against allowed datasources when URL comes from header
+        if use_header_auth and header_supplied_url:
+            allowed_datasources = _get_allowed_datasources_from_headers()
+            _validate_opensearch_url(opensearch_url, allowed_datasources)
+
         logger.info(
             f'Initializing single mode OpenSearch client for URL: '
             f'{_scrub_url_userinfo(opensearch_url)}'
@@ -672,6 +677,10 @@ def _initialize_client_multi_mode(cluster_info: ClusterInfo) -> AsyncOpenSearch:
                 raise AuthenticationError(
                     'Incomplete Basic credential in Authorization header: password is empty.'
                 )
+
+        # NOTE: No datasource validation needed in multi mode because the URL comes from
+        # trusted cluster configuration, not from request headers
+
         # Use common client creation function
         return _create_opensearch_client(
             opensearch_url=opensearch_url,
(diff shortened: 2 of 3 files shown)


vamsimanohar:
### Suggestion: one JSON header, and each datasource states its auth type

Thank you for the alternative design. This comment suggests a format for the same problem. It also answers question 4 in your list: yes, mixed credentials are necessary. A request must be able to use SigV4 for one cluster and basic auth for another.

The main point is this. A credential belongs to a datasource, not to the request. And a datasource must state which credential it uses. It must not receive one because a field is absent.

#### The problem to solve

A datasource is a complete configuration. It has a URL, a region, a service, and a credential. A format for many datasources must carry many complete configurations.

Both open proposals carry one credential for the whole request. #306 shares it by design. This PR can carry a credential for each datasource, but the code reads `opensearch-url-{i}` only, so the credential is still shared. In both cases a token for one cluster can go to a different cluster, and the model selects that cluster at the time of the tool call.

#### The format

Use one header with a JSON value. The names are the JSON keys, and each entry states its `auth_type`.

Two AWS datasources with one AWS credential:

```
aws-access-key-id: ASIA…
aws-secret-access-key: …
aws-session-token: …
opensearch-datasources: {"prod":{"auth_type":"aws_common_header","opensearch_url":"https://d1.us-west-2.es.amazonaws.com","aws_region":"us-west-2","aws_service_name":"es"},
                         "logs":{"auth_type":"aws_common_header","opensearch_url":"https://c1.us-east-1.aoss.amazonaws.com","aws_region":"us-east-1","aws_service_name":"aoss"}}
```

The credential appears one time. Each entry keeps its own region and service, so each request gets the correct signature.

Two clusters with different basic-auth users:

```
opensearch-datasources: {"alpha":{"auth_type":"basic_inline","opensearch_url":"https://alpha.internal:9200","opensearch_username":"ua","opensearch_password":"pa"},
                         "beta":{"auth_type":"basic_inline","opensearch_url":"https://beta.internal:9200","opensearch_username":"ub","opensearch_password":"pb"}}
```

One AWS domain and one self-managed cluster, in the same request:

```
aws-access-key-id: ASIA…
aws-secret-access-key: …
opensearch-datasources: {"aws-prod":{"auth_type":"aws_common_header","opensearch_url":"https://d1.us-west-2.es.amazonaws.com","aws_region":"us-west-2","aws_service_name":"es"},
                         "onprem":{"auth_type":"basic_inline","opensearch_url":"https://os.corp.internal:9200","opensearch_username":"u","opensearch_password":"p"}}
```

One datasource, with the headers that exist today. This does not change:

```
opensearch-url: https://d1.us-west-2.es.amazonaws.com
authorization: Bearer <token>
```

#### The four auth types

`aws_common_header` reads the AWS credential from the request headers, and takes the region and the service from the entry. It is valid for any number of datasources. `aws_inline`, `basic_inline` and `bearer_inline` hold the credential inside the entry. There is no `basic_common` type and no `bearer_common` type.

#### Why each datasource states its type

First, nothing happens by default. An operator reads the header and sees what each datasource uses.

Second, the server guesses nothing. `_create_opensearch_client` tries the credentials in this order: no auth, bearer, AWS keys, IAM role, basic. A check that asks "are AWS keys present?" gives the wrong answer when the request also carries a bearer token, because bearer comes first in that order. A stated type has no order to copy.

Third, and most important: a shared form exists only for a credential that never crosses the wire. SigV4 sends a signature. The host, the region, the service and the time are inside it, and the secret key stays on the server. So one AWS credential for many domains is safe, and it needs no repetition. A bearer token and a password go to the host without a change, so a second host receives a credential that it can use again. These two get no shared form. To use one password on two hosts, write it in both entries. The repetition is the statement of intent.

#### Rules for the header

`auth_type` has no default value. Each name must be unique, because the name is the selection key. `aws_region` is necessary for an AWS type; today a missing region reports *"No caller-supplied credentials for the requested URL"*, which points at the wrong problem. The header must reject `profile`, `iam_arn`, the certificate paths and `ssl_verify`, because a caller must not use the identity or the local files of the server. Any error must stop the whole request, and the server must never drop one entry and continue. Do not write this header to a log at any level.

#### About your question on header size

A header for each datasource gives more room inside each header. But the total limit for all the headers does not change, and the split costs about 25 percent more bytes, because each header repeats its name and each entry needs a `name` field that a JSON key gives at no cost. With a shared AWS credential an entry is about 130 bytes, so 25 datasources need 3.2 KB in one header. Neither limit is near.

One header also removes the index bookkeeping. The loop in this PR stops at the first gap, so `opensearch-url-0` and `opensearch-url-2` give a list with `-0` only, and the server writes no message. One header cannot lose an entry without a total failure.

The entry object is the same in both formats. So the framing can change later with no change to the entry schema, and a decision now costs nothing later.

#### On semantic names

Your trade-off list gives up the datasource names. The names are worth keeping, because they are what stops the model from handling a URL. With JSON keys the model calls `ListClustersTool`, gets `prod` and `logs`, and sends one name as `opensearch_cluster_name`. The server maps that name to the URL, the region, the service and the credential. The model never sees a URL, so it cannot request one.

PR #311, opened 2026-09-11
Filter and reject serverless-incompatible tools on AOSS connections

### LINKED ISSUE #310: Serverless-incompatible tools are advertised (and fail) against OpenSearch Serverless
## Summary

When the server is connected to an **OpenSearch Serverless (AOSS)** collection, it advertises and lets clients call tools that hit APIs Serverless does not implement (e.g. `_cluster/health`, `_cat/nodes`, `_stats`). These calls fail — against a real MCP/agent client they surface as errors or long hangs, and an LLM will often misattribute the failure to an unhealthy cluster. There is no serverless-aware tool filtering today, and the version-based gate that could have helped is silently disabled on Serverless.

## Root cause

1. **No serverless-aware filtering.** `core_tools` in `src/tools/tool_filter.py` is enabled by default and includes `ClusterHealthTool` and `GetShardsTool`. Nothing in the registry or `get_tools()` knows a connection is Serverless, so every catalog tool is offered regardless of backend.

2. **Version probe silently disables gating on Serverless.** `get_opensearch_version()` (`src/opensearch/helper.py`) probes via `client.info()` = `GET /`, which AOSS does not support. The exception is swallowed and the function returns `None`. `is_tool_compatible()` (`src/tools/utils.py`) then returns `True` unconditionally when the version is `None`, bypassing every `min_version`/`max_version` constraint and advertising the full catalog. The existing comment there — `# Find a version equivalent in serverless mode` — flags this as unfinished. In `multi` mode, `get_tools()` skips version filtering entirely, so tools are advertised unconditionally regardless of backend.

## Impacted tools

Endpoints checked against the [AOSS supported operations](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-genref.html). Tools that target an unsupported API:

| Tool | Endpoint | In default `core_tools`? |
|---|---|---|
| `ClusterHealthTool` | `GET /_cluster/health` | yes |
| `GetShardsTool` | `GET /_cat/shards` | yes |
| `GetClusterStateTool` | `GET /_cluster/state` | no |
| `GetSegmentsTool` | `GET /_cat/segments` | no |
| `CatNodesTool` | `GET /_cat/nodes` | no |
| `GetNodesTool` | `GET /_nodes` | no |
| `GetNodesHotThreadsTool` | `GET /_nodes/hot_threads` | no |
| `GetAllocationTool` | `GET /_cat/allocation` | no |
| `GetLongRunningTasksTool` | `GET /_cat/tasks` | no |
| `GetIndexStatsTool` | `GET /<index>/_stats` | no |
| `GetQueryInsightsTool` | `GET /_insights/top_queries` | no |

The non-core tools are only advertised when enabled via a category/regex/env, but because version gating is defeated (above) they are all offered on Serverless whenever enabled. Search-relevance and ML/experiment tools (`/_plugins/_search_relevance/*`) are likewise unsupported on Serverless but live in opt-in categories.

Tools verified **supported** on Serverless and correctly kept: `ListIndexTool` (`_cat/indices`, falls back to `_resolve/index`), `IndexMappingTool` (`_mapping`), `SearchIndexTool` (`_search`), `CountTool` (`_count`), `ExplainTool` (`_explain`), `MsearchTool` (`_msearch`), `PPLQueryTool` (`_plugins/_ppl`), `GetIndexInfoTool` (`GET /<index>`).

## Proposed fix

1. Add a serverless-incompatible tool set and filter it out when the connection is known to be Serverless — at list time in `single` mode (and `multi` mode when all configured clusters are serverless).
2. Add a **call-time guard** in the tool executor: if a call resolves to a Serverless connection and targets a serverless-incompatible tool, return a clear, immediate error instead of letting the request fail downstream. This covers header/per-request connection modes where the backend type is not known at list time.
3. Harden the version probe: when the version cannot be determined, log a clear warning rather than silently advertising the full catalog.

I'll open a PR implementing the above.

jiapingzeng:
### Description
Filter and reject serverless-incompatible tools on AOSS connections

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-mcp-server-py/issues/310


src/tools/tool_filter.py
```diff
@@ -21,6 +21,55 @@
 _resolved_allow_write_setting = None
 
 
+# Tools that target OpenSearch APIs which OpenSearch Serverless (AOSS) does not
+# implement (cluster, node and monitoring endpoints). They are filtered out at
+# list time when the connection is known to be serverless, and rejected at call
+# time otherwise. See:
+# https://docs.aws.amazon.com/opensearch-service/latest/developerguide/serverless-genref.html
+SERVERLESS_INCOMPATIBLE_TOOLS: frozenset = frozenset(
+    {
```
mingshl, line 48: how about ART agent's tools?They are not available on AOSS collections too?

src/tools/tool_filter.py
```diff
@@ -21,6 +21,55 @@
...
+        'GetClusterStateTool',  # GET /_cluster/state
+        'GetShardsTool',  # GET /_cat/shards
+        'GetSegmentsTool',  # GET /_cat/segments
+        'CatNodesTool',  # GET /_cat/nodes
+        'GetNodesTool',  # GET /_nodes
+        'GetNodesHotThreadsTool',  # GET /_nodes/hot_threads
+        'GetAllocationTool',  # GET /_cat/allocation
+        'GetLongRunningTasksTool',  # GET /_cat/tasks
+        'GetIndexStatsTool',  # GET /<index>/_stats
+        'GetQueryInsightsTool',  # GET /_insights/top_queries
```
mingshl: maybe a challenge here but how can we managed different group filter for the source of the truth. what api is available on AOSS or not,is it better to maintain a yaml file for tools available on AOS, and AOSS, then for the list and have a unit test to assert every new tools should be describe in the yaml config file. In this case, even AI would understand every tools should be considered added to supported in AOS and AOSS or not. just my two cents

mingshl:
This is a essential for filter out incompatible api for different datasource type as we support both AOS and AOSS data sources on mcp but some apis are not compatible. What is the expectation after this group filter in, what would be the agent response on some questions requiring the api but not available? 

would it be "sorry, i don't have the available tools to answer your question ". I am thinking the agent actually doesn't understand the mcp difference in connecting different data sources, but now the UI actually have auto suggestions for the next step, then i am curious on what would be the suggestions for these cases. have you test them? 

please add to this PR once you have a test in this case, and I want to see after this ship, what is the product experience and see if we need further enhancement

src/tools/tool_filter.py
jiapingzeng: Hmm, seems that there are more tools unsupported than supported. Perhaps a `SERVERLESS_COMPATIBLE_TOOLS` set would make more sense.

jiapingzeng:
> what would be the agent response on some questions requiring the api but not available?

The intention is that the agent would not see those tools to begin with. In single mode, for example, for classic OpenSearch it will have access to ClusterHealthTool, but not for Serverless. 

However, multi-mode with this approach would be a bit tricky. The current approach is to show all tools, but give the LLM an error if it tries to use an unsupported tool with serverless.

src/opensearch/client.py
```diff
@@ -40,6 +40,40 @@
...
+    """
+    if args is not None and getattr(args, 'aws_opensearch_serverless', None) is not None:
+        return bool(args.aws_opensearch_serverless)
+
+    try:
+        header_service = _get_auth_from_headers().get('aws_service_name')
+    except Exception:
+        header_service = None
+    if header_service:
+        return header_service.lower() == OPENSEARCH_SERVERLESS_SERVICE
```
nathaliellenaa: #306 makes `aws-service-name` comma-separated (`es,aoss,...`), so this equality check would return `False` for any multi-datasource request and serverless would go undetected. Should we split the value here?

src/opensearch/client.py
```diff
@@ -40,6 +40,40 @@
...
+
+    Mirrors the precedence used when building the client: explicit per-call flag,
+    per-request aws-service-name header, configured multi-mode cluster, then
+    environment flag and URL heuristic.
+    """
+    if args is not None and getattr(args, 'aws_opensearch_serverless', None) is not None:
+        return bool(args.aws_opensearch_serverless)
+
+    try:
+        header_service = _get_auth_from_headers().get('aws_service_name')
```
nathaliellenaa: Should we gate this on header auth being enabled ([example](https://github.com/opensearch-project/opensearch-mcp-server-py/blob/main/src/opensearch/client.py#L459-#L461))?

src/tools/tool_filter.py
mingshl: I am thinking having a file tag all tools, each with SERVERLESS_COMPATIBLE_TOOLS might be better, My concern is someone added a new tool without thinking about is it severless compatible, then it's hard to maintain

nathaliellenaa:
>However, multi-mode with this approach would be a bit tricky. The current approach is to show all tools, but give the LLM an error if it tries to use an unsupported tool with serverless.

Adding to this, I think multi model could be handled more properly with dynamic tool filtering. That way, the tool list is computed per request based on the service actually in play rather than once at startup. We can keep this as a follow up

nathaliellenaa (approved):
Thanks @jiapingzeng , LGTM

